PR #1, opened 2022-03-21
ROE-55 initial skeleton api

markpit:
Creating an initial skeleton api
* added a post endpoint on the application root that will return a string
* added the spring boot actuator which gives us the /actuator/health endpoint
* added routes.yaml
* added Makefile
* added start.sh
* added version file
* added the latest structured logging and the latest log4j (structured logging is a minor version behind log4j latest)


routes.yaml
```diff
@@ -0,0 +1,6 @@
+app_name: overseas-entities-api
+group: api
+
+weight: 90
+routes:
+ 1: ^/register-an-overseas-entity/*
```
mouhajer-ch, line 6: Does create a conflict with web endpoints?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/OverseasEntitiesApiApplication.java
```diff
@@ -0,0 +1,22 @@
...
+
+@SpringBootApplication
+public class OverseasEntitiesApiApplication {
+
+    public static void main(String[] args) {
+        SpringApplication.run(OverseasEntitiesApiApplication.class, args);
+    }
+
+    @PostConstruct
+    public void init() {
```
mouhajer-ch, line 17: Do we need this method now?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.companieshouse.overseasentitiesapi.controller;
```
mouhajer-ch, line 1: Probably we can use an `HealthcheckController` instead

pom.xml
```diff
@@ -0,0 +1,118 @@
...
+        <groupId>uk.gov.companieshouse</groupId>
+        <artifactId>companies-house-spring-boot-parent</artifactId>
+        <version>1.5.0-rc1</version> <!-- lookup parent from repository -->
+    </parent>
+    <artifactId>overseas-entities-api</artifactId>
+    <version>unversioned</version>
+    <name>overseas-entities-api</name>
+    <description>API for register an overseas entity service</description>
+    <properties>
+        <java.version>11</java.version>
```
mouhajer-ch, line 15: I though we are on `1.8`?

mouhajer-ch (commented):
unit test are missing, I would have included in the first skeleton

routes.yaml
markpit: I followed the example from confirmation-statement-web and api, and those used the weighting to avoid clash, but as you said, the issue could have possibly been fixed by removing the matching url from the api routes.yaml. If I remove this endpoint, we'll have an empty routes.yaml, unless i change it to another value, but not sure what...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/OverseasEntitiesApiApplication.java
markpit: strictly speaking no, but figured if I removed it then it would be forgotten about and we'll end up having date/time values being out of sync between mongo and the api, which might cause us time to investigate and resolve.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
markpit: the inclusion of the spring-boot-starter-actuator in the pom will automatically create a /actuator/health endpoint for the app that will return the health status.
I figured we would at least need a minimum of a post endpoint, or at least something to prove our controller was working.

pom.xml
markpit: We should be safe to use 11, had a quick look and insolvency-data-api is also using 11 - https://github.com/companieshouse/insolvency-data-api/blob/main/pom.xml#L19
Also our confirmation-statement-api is running  11 - https://github.com/companieshouse/confirmation-statement-api/blob/main/pom.xml#L15

markpit:
> unit test are missing, I would have included in the first skeleton

I'll add a test to cover the controller

routes.yaml
mouhajer-ch: We are going to have a transaction based api, we can use the same syntax as confirmation, something like  `1: ^/transactions/(.*)/register-an-overseas-entity/*`

src/main/java/uk/gov/companieshouse/overseasentitiesapi/OverseasEntitiesApiApplication.java
mouhajer-ch: Awesome, thanks for the update

pom.xml
mouhajer-ch: Good,I will change the version on my local profile config to work with the 11

routes.yaml
mouhajer-ch: We can have a look at it in another story

mouhajer-ch (approved):

mwestacott (approved):

PR #12, opened 2022-04-03
ROE-148 Updating api spec to include entity and presenter data

markpit:
Updating schema.json to include entity and presenter data and update endpoint url to include transaction id

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -11,18 +11,36 @@
   ],
   "tags": [
     {
-      "name": "Overseas Entities"
+      "name": "Overseas Entity"
     }
   ],
   "paths": {
-    "/overseas-entities": {
+    "/transaction/{transaction_id}/overseas-entity/": {
+      "parameters": [
+        {
+          "name": "transaction_id",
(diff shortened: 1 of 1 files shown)


spec/schema.json
```diff
@@ -33,5 +51,131 @@
...
+        "registrationNumber"
+      ]
+    },
+    "Address": {
+      "type": "object",
+      "properties": {
+        "propertyNameNumber": {
+          "type": "string"
+        },
+        "addressLine1": {
```
mwestacott, line 154: Just line 1 and line 2 here - fields are in an 'address' object and you haven't prefixed the other fields with 'address'.

mouhajer-ch (commented):
Hi, entity and presenter fields will not match what it is in the web. We need to create a map or update them, I do not mind to approve it but I would like to have a quick call.

mwestacott (approved):

mwestacott (changes requested):

mwestacott (approved):

PR #13, opened 2022-04-04
Roe 149 add submission model

mwestacott:
* Model classes added and annotated with mappings to JSON fields
* Parent model class added to controller POST method as a request
body parameter


src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/OverseasEntitySubmission.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.companieshouse.overseasentitiesapi.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Map;
```
markpit, line 5: is the map used?

mwestacott: No. Good spot. Will remove.

markpit (approved):

PR #17, opened 2022-04-05
ROE-151 add transaction interceptor

markpit:
https://companieshouse.atlassian.net/browse/ROE-151

Adding transactionInterceptor to perform security checks via transaction-api 'get' and to supply the controller with the transaction specified in the url.
Also updated the endpoint path to include 'transactions'

branch name was incorrectly named, should have been roe-151

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/configuration/InterceptorConfig.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/configuration/InterceptorConfig.java
@@ -1,0 +1,36 @@
+package uk.gov.companieshouse.overseasentitiesapi.configuration;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.lang.NonNull;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+import uk.gov.companieshouse.overseasentitiesapi.interceptor.TransactionInterceptor;
+
+@Configuration
+public class InterceptorConfig implements WebMvcConfigurer {
+
+    static final String TRANSACTIONS = "/transactions/**";
+
+    @Autowired
+    private TransactionInterceptor transactionInterceptor;
+
+    /**
+     * Setup the interceptors to run against endpoints when the endpoints are called
+     * Interceptors are executed in the order they are added to the registry
+     * @param registry the spring interceptor registry
+     */
+    @Override
+    public void addInterceptors(@NonNull InterceptorRegistry registry) {
+        addTransactionInterceptor(registry);
+    }
+
+    /**
+     *  Interceptor to get transaction and put in request for endpoints that require a transaction
+     * @param registry the spring interceptor registry
+     */
+    private void addTransactionInterceptor(InterceptorRegistry registry) {
+        registry.addInterceptor(transactionInterceptor)
+                .addPathPatterns(TRANSACTIONS);
+    }
+}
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/TransactionInterceptor.java
```diff
@@ -0,0 +1,54 @@
...
+        String passthroughHeader = request.getHeader(ApiSdkManager.getEricPassthroughTokenHeader());
+
+        var logMap = new HashMap<String, Object>();
+        logMap.put(TRANSACTION_ID_KEY, transactionId);
+        String reqId = request.getHeader(ERIC_REQUEST_ID_KEY);
+        try {
+            ApiLogger.debugContext(reqId, "Getting transaction for request.", logMap);
+
+            final var transaction = transactionService.getTransaction(transactionId, passthroughHeader);
+            ApiLogger.debugContext(reqId, "Transaction retrieved.", logMap);
```
mwestacott, line 44: Should we add the retrieved transaction to the 'logMap' before this call to 'debug'?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
```diff
@@ -24,6 +26,7 @@ public OverseasEntitiesController(OverseasEntitiesService overseasEntitiesServic
 
     @PostMapping("/")
     public ResponseEntity<String> createNewSubmission(
+            @RequestAttribute("transaction") Transaction transaction,
             @RequestBody OverseasEntitySubmission overseasEntitySubmission) {
         ApiLogger.debug("Called createNewSubmission()");
```
mwestacott, line 32: This change is fine, but we need to link the transaction to the new overseas entity submission (probably in the service call). We don't currently have a JIRA sub-task to do this. I'll add one if you agree @markpit ?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/configuration/InterceptorConfig.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.lang.NonNull;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+import uk.gov.companieshouse.overseasentitiesapi.interceptor.TransactionInterceptor;
+
+@Configuration
+public class InterceptorConfig implements WebMvcConfigurer {
```
mwestacott, line 11: Are you missing a 'component scan' annotation here?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
markpit: yes we'll need to do it at some point, can be added as new task

src/main/java/uk/gov/companieshouse/overseasentitiesapi/configuration/InterceptorConfig.java
markpit: don't think so, the scan in confirmation-statement was because we were importing an interceptor from a pom dependency that had a different package structure, so we had to use ComponentScan and specify it's path to pick it up. We'll have to add that interceptor at some point as it is security role related. It's working as @Configuartion is also a @Component so can see other beans in the project.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/TransactionInterceptor.java
markpit: I wouldn't say put the whole transaction in as I don't think we should log it out, perhaps adding the transaction_id might be better so we know what transaction was retrieved?

mwestacott: Ok. Transaction id already logged added/logged on line 38 so I think we're good here.

mwestacott (approved):

PR #19, opened 2022-04-07
ROE-150 Separate the DTO and DAO models

mwestacott:
* Implementation in line with recommendation from Sonar
* DTO (JSON) model classes now added
* DAO (Mongo) model classes also added
* Mapper called from service to move JSON data into Mongo object/format
* Unit test added


pom.xml
```diff
@@ -17,6 +17,7 @@
         <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
         <structured-logging.version>1.9.14</structured-logging.version>
         <log4j.version>2.17.2</log4j.version>
+        <org.mapstruct.version>1.5.0.RC1</org.mapstruct.version>
```
mwestacott, line 20: We used version 1.4.2 on Confirmation Statement but used the latest version available here.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/EntityDto.java
```diff
@@ -1,63 +1,41 @@
...
+    @JsonProperty("legal_form")
     private String legalForm;
 
-    @JsonProperty(LAW_GOVERNED_FIELD)
-    @Field(LAW_GOVERNED_FIELD)
+    @JsonProperty("law_governed")
     private String lawGoverned;
 
-    @JsonProperty(PUBLIC_REGISTER_ENTITY_REGISTERED_ON_FIELD)
+    @JsonProperty("public_register_entity_registered_on")
```
markpit, line 34: public_register_entity_registered_on is not meant to be a date, it is just known as 'publicRegister' in the web app - 'On which public register is the overseas entity registered'
probably needs renaming

markpit, line 34: public_register_entity_registered_on is not meant to be a date, it is just known as 'publicRegister' in the web app - 'On which public register is the overseas entity registered'
probably needs renaming

mwestacott: Duplicate comment? Will resolve this convo.

mwestacott: Fixed. Updated the spec with the changed field name too.

markpit (approved):

PR #27, opened 2022-04-12
Fix - adding jacoco version to try and get test coverage working

markpit:
Sonar is not reporting unit test coverage, possibly caused by jacoco version missing from pom. Adding it to try and resolve issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,6 +143,7 @@
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
+                <version>${jacoco-maven-plugin.version}</version>
             </plugin>
         </plugins>
     </build>


pom.xml
```diff
@@ -143,6 +143,7 @@
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
+                <version>${jacoco-maven-plugin.version}</version>
```
mwestacott, line 146: Doesn't version need defining? Like for maven-surefire-plugin.version?

markpit: it seems to be populated (0.8.4) when hovering over in intellij, assumed it is imported from parent pom, i can check

mwestacott: Will assume it's good and approve. We'll need to merge anyway to see if it fixes things.

mwestacott (approved):

PR #33, opened 2022-04-14
ROE-256 add beneficial owner spec

markpit:
https://companieshouse.atlassian.net/browse/ROE-256

Updating schema spec to include the data structure for beneficial owners and managing officers

The nature of control options on the V4 prototype are check boxes where you can select more than one, so have set the nature of controls to be arrays.

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -243,7 +243,7 @@
         "legal_form": {
           "type": "string"
         },
-        "governing_law": {
+        "law_governed": {
           "type": "string"
         },
         "public_register_name": {
@@ -335,7 +335,7 @@
         "legal_form": {
           "type": "string"
         },
-        "governing_law": {
+        "law_governed": {
           "type": "string"
         },
         "is_on_register_in_country_formed_in": {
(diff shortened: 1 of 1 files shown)


mwestacott (approved):

spec/schema.json
```diff
@@ -213,7 +243,7 @@
         "legal_form": {
           "type": "string"
         },
-        "law_governed": {
+        "governing_law": {
```
mwestacott, line 246: Oh, just a thought on this - it will mean a change in the API DTO model class too. And in the chips-filing-consumer model that I'm currently working on. We'll have to be careful with these renames/changes on any objects that we've mapped in the API and filing consumer...

markpit: ok, i'm happy to change it back to law_governed

mwestacott: I don't mind the change, just need to be sure not to forget to change this elsewhere.

mouhajer-ch: If you change it we need to change on line 233

mwestacott (approved):

mouhajer-ch (approved):

PR #44, opened 2022-05-05
ROE-168 Keep entity fields up to date

mattch1:
https://companieshouse.atlassian.net/browse/ROE-398

Updated field: "public_register_name" to "register_name"
Added field: isOnRegisterInCountryFormedIn

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -249,7 +249,7 @@
         "law_governed": {
           "type": "string"
         },
-        "register_name": {
+        "public_register_name": {
           "type": "string"
         },
         "registration_number": {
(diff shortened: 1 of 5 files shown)


spec/schema.json
```diff
@@ -249,11 +249,14 @@
         "law_governed": {
           "type": "string"
         },
-        "public_register_name": {
+        "register_name": {
```
mwestacott, line 252: Changing this is fine, but it will have a knock on effect in 'chips-filing-consumer' which has already mapped this data and is expecting a field named 'public_register_name'. I can't remember, but it may even fall over if the field name changes.

markpit: Also, the Web entity model has been built using the public_register_name so that would also need changing, probably best to leave as public_register_name

mwestacott: Corporate BOs aren't implemented yet in `chips-filing-consumer` so changes there are still ok. If web, api and spec are all in sync then this should be fine.

PR #45, opened 2022-05-06
ROE-429 Submit BO Individual data to CHIPS

mwestacott:
* Additional data now returned from OE filings end-point
* Unit test added

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
@@ -8,16 +8,19 @@ import java.util.Map;
 
 public class OverseasEntitySubmissionDto {
 
+    public static final String ENTITY_FIELD = "entity";
+    public static final String BENEFICIAL_OWNERS_INDIVIDUAL_FIELD = "beneficial_owners_individual";
+
     @JsonProperty("presenter")
     private PresenterDto presenter;
 
-    @JsonProperty("entity")
+    @JsonProperty(ENTITY_FIELD)
     private EntityDto entity;
 
     @JsonProperty("beneficial_owners_statement")
     private BeneficialOwnersStatementType beneficialOwnersStatement;
 
-    @JsonProperty("beneficial_owners_individual")
+    @JsonProperty(BENEFICIAL_OWNERS_INDIVIDUAL_FIELD)
     private List<BeneficialOwnerIndividualDto> beneficialOwnersIndividual;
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
```diff
@@ -8,16 +8,19 @@
 
 public class OverseasEntitySubmissionDto {
 
+    public static final String ENTITY_FIELD = "entity";
```
mwestacott, line 11: I defined these as the exact JSON name string is also used in the `FilingsService` class and I didn't like the duplication. What do we think?

Also, in an agile fashion, I've only defined constants for the fields currently going to CHIPS but could define them all now if that's the preference.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
```diff
@@ -8,16 +8,19 @@
...
+    public static final String BENEFICIAL_OWNERS_INDIVIDUAL_FIELD = "beneficial_owners_individual";
+
     @JsonProperty("presenter")
     private PresenterDto presenter;
 
-    @JsonProperty("entity")
```
mattch1, line 14: what : entity_field
why: concern: will need to amend the stack to get it to agree with the new name
where: just a question really: is there any way of qualifying it when it goes to the filing service?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
markpit: seems ok to me, the filing service will just reference the constant in the dto to ensure the json fields are named the same when exporting from /filings endpoint. The json properties naming will remain unchanged, just referencing a constant to avoid typos etc..  Happy to convert them as we go.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
mwestacott: @mattch1 no need to 'amend the stack' (do you mean spec, web and other model classes?) because name hasn't actually changed. It's still "entity" but now just referencing a constant.

Does that answer your question?

markpit (approved):

mwestacott:
Can now be merged.

PR #47, opened 2022-05-09
ROE-449 Ensure register name has public prefixed

mattch1:
https://companieshouse.atlassian.net/browse/ROE-449

Updated "register_name" to "public_register_name" in corporate bo and would have done for corporate maaging officer but this is not present yet so updated it in the spec only.

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -348,7 +348,7 @@
         "is_on_register_in_country_formed_in": {
           "type": "boolean"
         },
-        "register_name": {
+        "public_register_name": {
           "type": "string"
         },
         "registration_number": {
@@ -486,7 +486,7 @@
         "is_on_register_in_country_formed_in": {
           "type": "boolean"
         },
-        "register_name": {
+        "public_register_name": {
           "type": "string"
         },
         "registration_number": {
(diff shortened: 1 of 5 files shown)


mwestacott:
Looks fine but presume we need to coordinate merging this with the corresponding change in the web/sdk project. Is that one ready to go to?

spec/schema.json
```diff
@@ -486,7 +486,7 @@
         "is_on_register_in_country_formed_in": {
           "type": "boolean"
         },
-        "register_name": {
+        "public_register_name": {
```
mouhajer-ch, line 489: BeneficialOwnerGovernmentOrPublicAuthority does not have `public_register_name`, I think can be removed with `is_on_register_in_country_formed_in` and `registration_number`

mattch1: This is for ManagingOfficerCorporate

mouhajer-ch: Ops, you are right Mark! sorry mates

markpit: We'll have to look at BeneficialOwnerGovernmentOrPublicAuthority separately in another PR, I can see that there is no register_name or public_register_name in the spec for it, but there is also no field for it currently in the web model, but there appears to be a field for it in the v5 prototype.

mouhajer-ch: Nope, the register fields are on `beneficial-owner-other`, `managing-officer-corporate` and `entity`

markpit: Yes, you are right, I was looking at the corporate page! too confusing by half!

markpit (approved):

PR #53, opened 2022-05-12
ROE-554 Adding /costs endpoint

markpit:
https://companieshouse.atlassian.net/browse/ROE-554

Adding /costs endpoint 
* Updated api-sdk-java version to latest

### DIFF
--- a/src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/CostsServiceTest.java
+++ b/src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/CostsServiceTest.java
@@ -1,0 +1,40 @@
+package uk.gov.companieshouse.overseasentitiesapi.service;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.test.util.ReflectionTestUtils;
+
+import java.util.Collections;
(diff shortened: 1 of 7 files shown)


mwestacott (approved):

src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/CostsServiceTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+@ExtendWith(MockitoExtension.class)
+class CostsServiceTest {
+
+    @InjectMocks
+    private CostsService costService;
+
+    @BeforeEach
+    void init() {
+        ReflectionTestUtils.setField(costService, "costAmount", "13.00");
```
mouhajer-ch, line 23: Quick one, probably it should be the `OE01_COST` value instead of embedded `13.00`

markpit: I'm not sure we'd have the env vars setup in the concourse environment that runs the tests. It shouldn't matter in this scenario as we are setting the costAmount field in the costService to amount x, and the test checks that the same amount x is the cost returned in the Cost object from the method.

mouhajer-ch: Sure, it is not a blocker I was thinking something like `private static final String OE01_COST = "13.00";` and reusing it on the test. Anyway it is not a blocker, I will approve it

mouhajer-ch (approved):

PR #56, opened 2022-05-16
ROE-78 Update costs endpoint and feature flag payments

markpit:
https://companieshouse.atlassian.net/browse/ROE-78

* Add logging to costs endpoint
* Add feature flag to enable payment integration (or not)
* removed 'variable payments' from costs data, not used in confirmation statement and can't see it used in the transactions-api so not sure what it is, safer to remove if payments are ok in Confirmation statement like that


mouhajer-ch (approved):
It looks good to me, Tested `e2e` works as expected

PR #69, opened 2022-06-22
Map address data from DAO back to DTO for filings endpoint

jphillips2121:
Added trust data to the filings endpoint and maps the DAO to the DTO in the correct address format that CHIPS is expecting for `individual` and `corporate` trust information. 

Resolves: ROE-656

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/trust/TrustCorporateDto.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/trust/TrustCorporateDto.java
@@ -1,11 +1,13 @@
 package uk.gov.companieshouse.overseasentitiesapi.model.dto.trust;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.AddressDto;
 
 import java.time.LocalDate;
 
 public class TrustCorporateDto {
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("type")
     private String type;
 
@@ -15,57 +17,81 @@ public class TrustCorporateDto {
     @JsonProperty("date_became_interested_person")
     private LocalDate dateBecameInterestedPerson;
 
+    @JsonProperty("registered_office_address")
+    private AddressDto registeredOfficeAddress;
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_line_1")
     private String roAddressLine1;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_line_2")
     private String roAddressLine2;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_care_of")
     private String roAddressCareOf;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_country")
     private String roAddressCountry;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_locality")
     private String roAddressLocality;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_po_box")
     private String roAddressPoBox;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_postal_code")
     private String roAddressPostalCode;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_premises")
     private String roAddressPremises;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_region")
     private String roAddressRegion;
 
+    @JsonProperty("service_address")
+    private AddressDto serviceAddress;
+
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_line_1")
     private String saAddressLine1;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_line_2")
     private String saAddressLine2;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_care_of")
     private String saAddressCareOf;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_country")
     private String saAddressCountry;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_locality")
     private String saAddressLocality;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_po_box")
     private String saAddressPoBox;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_postal_code")
     private String saAddressPostalCode;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_premises")
     private String saAddressPremises;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("sa_address_region")
     private String saAddressRegion;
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -96,6 +101,14 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
         data.put(MANAGING_OFFICERS_INDIVIDUAL_FIELD, submissionDto.getManagingOfficersIndividual());
         data.put(MANAGING_OFFICERS_CORPORATE_FIELD, submissionDto.getManagingOfficersCorporate());
         data.put(BENEFICIAL_OWNERS_STATEMENT, submissionDto.getBeneficialOwnersStatement());
+
+        // Convert trust data to JSON string
+        ObjectMapper mapper = JsonMapper.builder().findAndAddModules().build();
+        try {
+            data.put(TRUST_DATA, mapper.writeValueAsString(submissionDto.getTrusts()));
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
markpit, line 110: I wouldn't use printStackTrace here, could you throw a ServiceException, you should be able to use the e.getMessage as message and pass the original exception e in as another param to ServiceException

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -1,5 +1,9 @@
 package uk.gov.companieshouse.overseasentitiesapi.service;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
```
markpit, line 5: unused import here

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -96,6 +101,14 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
         data.put(MANAGING_OFFICERS_INDIVIDUAL_FIELD, submissionDto.getManagingOfficersIndividual());
         data.put(MANAGING_OFFICERS_CORPORATE_FIELD, submissionDto.getManagingOfficersCorporate());
         data.put(BENEFICIAL_OWNERS_STATEMENT, submissionDto.getBeneficialOwnersStatement());
+
+        // Convert trust data to JSON string
+        ObjectMapper mapper = JsonMapper.builder().findAndAddModules().build();
```
markpit, line 106: could you add a test to the FilingsServiceTest to test that the  submissionDto.getTrusts() get's converted to a string correctly?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/trust/TrustCorporateDto.java
```diff
@@ -267,19 +300,26 @@ public void setSaAddressRegion(String saAddressRegion) {
...
-        serviceAddress.setLine1(saAddressLine1);
-        serviceAddress.setLine2(saAddressLine2);
-        serviceAddress.setCounty(saAddressRegion);
-        serviceAddress.setLocality(saAddressLocality);
-        serviceAddress.setCountry(saAddressCountry);
-        serviceAddress.setCareOf(saAddressCareOf);
-        serviceAddress.setPoBox(saAddressPoBox);
-        serviceAddress.setPostcode(saAddressPostalCode);
+        // If the address is being set from the sa fields generate the address
+        if (saAddressPostalCode != null) {
```
mwestacott, line 304: Please see comment above, but would also like to double-check this is correct - could principal or service address details be present without postcodes?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/trust/TrustCorporateDto.java
```diff
@@ -15,57 +17,81 @@ public class TrustCorporateDto {
...
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_premises")
     private String roAddressPremises;
 
+    @JsonInclude(JsonInclude.Include.NON_NULL)
     @JsonProperty("ro_address_region")
     private String roAddressRegion;
 
+    @JsonProperty("service_address")
```
mwestacott, line 59: Wondering why we add a 'service address' object as well as all the individual address fields?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
mwestacott: Yes. It should be checking that those NON_NULL definitions are working (and when not present, null is output).

Also need a test when no trust data is present.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/trust/TrustCorporateDto.java
```diff
@@ -181,17 +210,24 @@ public void setRoAddressRegion(String roAddressRegion) {
...
-        address.setCountry(roAddressCountry);
-        address.setCareOf(roAddressCareOf);
-        address.setPoBox(roAddressPoBox);
-        address.setPostcode(roAddressPostalCode);
-        return address;
+        // If the address is being set from the ro fields generate the address
+        if (Objects.nonNull(roAddressPostalCode)) {
+            registeredOfficeAddress = new AddressDto();
+            registeredOfficeAddress.setPropertyNameNumber(saAddressPremises);
+            registeredOfficeAddress.setLine1(saAddressLine1);
```
mwestacott, line 217: Should set roAddress fields...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/trust/TrustIndividualDto.java
```diff
@@ -288,17 +324,24 @@ public void setUraAddressRegion(String uraAddressRegion) {
...
-        address.setLine1(uraAddressLine1);
-        address.setLine2(uraAddressLine2);
-        address.setCounty(uraAddressRegion);
-        address.setLocality(uraAddressLocality);
-        address.setCountry(uraAddressCountry);
-        address.setCareOf(uraAddressCareOf);
-        address.setPoBox(uraAddressPoBox);
-        address.setPostcode(uraAddressPostalCode);
-        return address;
+        // If the address is being set from the ura fields generate the address
```
mwestacott, line 327: Suggest making this comment a bit more explicit, as discussed. E.g.

"When converting from DTO to DAO the individual address fields will be present and need to be converted to an address object during the mapping process. When converting from DAO to DTO the individual fields will not be populated and so the address object just needs to be returned"

Probably then just need to duplicate this comment in all the places that this is happening. Four?

As we all agreed, a better and consistent interface between web and API for the trust data would be the best fix for this...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/trust/TrustCorporateDto.java
```diff
@@ -1,71 +1,100 @@
...
+    @JsonInclude(NON_NULL)
     @JsonProperty("type")
     private String type;
 
     @JsonProperty("name")
     private String name;
 
-    @JsonProperty("date_became_interested_person")
+    @JsonInclude(NON_NULL)
     private LocalDate dateBecameInterestedPerson;
```
mwestacott, line 21: Just wondering why these two fields don't have JsonProperty annotations on them (with snake_case names)?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -96,6 +100,19 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
         data.put(MANAGING_OFFICERS_INDIVIDUAL_FIELD, submissionDto.getManagingOfficersIndividual());
         data.put(MANAGING_OFFICERS_CORPORATE_FIELD, submissionDto.getManagingOfficersCorporate());
         data.put(BENEFICIAL_OWNERS_STATEMENT, submissionDto.getBeneficialOwnersStatement());
+
+        if (submissionDto.getTrusts() != null) {
```
mwestacott, line 104: I would use Objects.nonNull() here for the check.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -96,6 +100,19 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
...
+            // Convert trust data to JSON string if it exists on transaction else set it to an empty string
+            ObjectMapper mapper = JsonMapper.builder().findAndAddModules().build();
+            try {
+                data.put(TRUST_DATA, mapper.writeValueAsString(submissionDto.getTrusts()));
+            } catch (JsonProcessingException e) {
+                throw new ServiceException("Error converting trust data to JSON " + e.getMessage(), e);
+            }
+        } else {
```
mwestacott, line 112: I'd write this slightly differently, but up to you... Would define a trustData string variable before the if block and initialise it as an empty string - code in if block would overwrite its value. The else block can then go. And you'd only have one data.put() line immediately after the if block that sets the 'trustData' variable on the data map.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingServiceTest.java
```diff
@@ -133,7 +133,36 @@ void initGetPaymentMocks() throws ApiErrorResponseException, URIValidationExcept
...
+        assertEquals(FILING_KIND_OVERSEAS_ENTITY, filing.getKind());
+        assertEquals(FILING_DESCRIPTION_IDENTIFIER, filing.getDescriptionIdentifier());
+        assertEquals("Filing Description with registration date 26 March 2022", filing.getDescription());
+        final PresenterDto presenterInFiling = (PresenterDto)filing.getData().get("presenter");
+        assertEquals("Joe Bloggs", presenterInFiling.getFullName());
+        assertEquals("user@domain.roe", presenterInFiling.getEmail());
+        final EntityDto entityInFiling = ((EntityDto) filing.getData().get("entity"));
+        assertEquals("Joe Bloggs Ltd", entityInFiling.getName());
+        assertEquals("Eutopia", entityInFiling.getIncorporationCountry());
+        assertTrue((filing.getData().get("trusts")) instanceof String);
```
mwestacott, line 158: I don't think this assertion alone is enough. Would additionally check that data.trusts value contains a JSON string with the values you set in the mock (above).

mwestacott (approved):

PR #72, opened 2022-07-04
Add trust json information to individual BOs

jphillips2121:
When calling the `/filing` endpoint loop through the `trust_ids` and return individual trust data on each BO rather than a whole block of JSON data for the entire submission

Also remove trust data for government or public authorities

Resolves: ROE-889

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 import uk.gov.companieshouse.api.handler.exception.URIValidationException;
 import uk.gov.companieshouse.api.model.filinggenerator.FilingApi;


mwestacott:
Please could you get a JIRA ticket reference into the title and/or description please.

jphillips2121:
> Please could you get a JIRA ticket reference into the title and/or description please.

There is one in the description already

spec/schema.json
```diff
@@ -412,7 +413,8 @@
         "trust_ids": {
           "type": "array",
           "items": {
-            "type": "string"
+            "type": "string",
```
mwestacott, line 416: Minor, but suggest making one of these examples what the value looks like when it's an array of ids. Is that comma-separated? Not sure...

mwestacott:
> > Please could you get a JIRA ticket reference into the title and/or description please.
> 
> There is one in the description already

DOH Sorry! See it now.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -95,26 +99,120 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
...
+                            throw new ServiceException("No trusts exist for this filing but a trust id is provided for BO Individual "
+                                    + beneficialOwner.getFirstName() + " " + beneficialOwner.getLastName());
+                        }
+                        for (TrustDataDto trust : submissionDto.getTrusts()) {
+                            if (trust.getTrustId().equals(trustId)) {
+                                trustData.add(trust);
+                            }
+                        }
+
+                        // If there is more than one trust with the same ID throw an error
```
mwestacott, line 133: I think this check is ensuring that a beneficial owner can only be linked to one trust? If yes, suggest rephrasing the comment and exception message just to make this a little clearer.

I might be wrong though, because if that's the case, why do we support multiple trust ids per BO?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -95,26 +99,120 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
...
+                        List<TrustDataDto> trustData = new ArrayList<>();
+
+                        if (!Objects.nonNull(submissionDto.getTrusts())) {
+                            throw new ServiceException("No trusts exist for this filing but a trust id is provided for BO Individual "
+                                    + beneficialOwner.getFirstName() + " " + beneficialOwner.getLastName());
+                        }
+                        for (TrustDataDto trust : submissionDto.getTrusts()) {
+                            if (trust.getTrustId().equals(trustId)) {
+                                trustData.add(trust);
+                            }
```
mwestacott, line 130: I guess we could 'break' out of the loop at this point (but then we wouldn't pick up on the invalid mappings). So this is probably fine.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -95,26 +99,120 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
...
+                        }
+                        // If there are is no trust with that ID throw an error
+                        if (trustData.isEmpty()) {
+                            throw new ServiceException("There are no trusts for the ID: " + trustId);
+                        }
+
+                        trustsDataForBO.add(trustData.get(0));
+                    }
+                }
+
```
mwestacott, line 145: There seems to be two logical parts to this (quite big) method - the bit above that finds the trust data for BOs and the next bit that converts it to a JSON string. Suggest splitting this method up into a couple of private methods that do the two bits of processing Will improve readability.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -95,26 +99,120 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
...
 
-        ApiLogger.debug("Submission data has been set on filing", logMap);
+        return beneficialOwnersIndividualSubmissionData;
+    }
+
+    private List<BeneficialOwnerCorporateDto> getBeneficialOwnersCorporateSubmissionData(OverseasEntitySubmissionDto submissionDto) throws ServiceException {
+        List<BeneficialOwnerCorporateDto> beneficialOwnersCorporateSubmissionData = new ArrayList<>();
+
+        if (Objects.nonNull(submissionDto.getBeneficialOwnersCorporate())) {
+            for (BeneficialOwnerCorporateDto beneficialOwner : submissionDto.getBeneficialOwnersCorporate()) {
```
mwestacott, line 169: Parts of the body of this method are more or less an exact match of the above method. If you're familiar with lambda functions and passing them in as method parameters you could refactor to have a common method. This might become clearer after doing the method split suggest above.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -95,26 +99,120 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
...
-                trustData = mapper.writeValueAsString(submissionDto.getTrusts());
-            } catch (JsonProcessingException e) {
-                throw new ServiceException("Error converting trust data to JSON " + e.getMessage(), e);
+        ApiLogger.debug("Submission data has been set on filing", logMap);
+    }
+
+    private List<BeneficialOwnerIndividualDto> getBeneficialOwnersIndividualSubmissionData(OverseasEntitySubmissionDto submissionDto) throws ServiceException {
+        List<BeneficialOwnerIndividualDto> beneficialOwnersIndividualSubmissionData = new ArrayList<>();
+
+        if (Objects.nonNull(submissionDto.getBeneficialOwnersIndividual())) {
```
markpit, line 115: I would use a guard clause to return early if it is null- 
 if (Objects.isNull(submissionDto.getBeneficialOwnersIndividual())) {
    return beneficialOwnersCorporateSubmissionData;
}
that way you can then remove a layer of 'wrapping'/indentaion from the outer for loop

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -95,26 +99,120 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
...
+
+        if (Objects.nonNull(submissionDto.getBeneficialOwnersCorporate())) {
+            for (BeneficialOwnerCorporateDto beneficialOwner : submissionDto.getBeneficialOwnersCorporate()) {
+                List<TrustDataDto> trustsDataForBO = new ArrayList<>();
+                // Loop through each trustId(s) and lookup the information for that trust
+                if (Objects.nonNull(beneficialOwner.getTrustIds())) {
+                    for (String trustId : beneficialOwner.getTrustIds()) {
+                        List<TrustDataDto> trustData = new ArrayList<>();
+
+                        if (!Objects.nonNull(submissionDto.getTrusts())) {
```
markpit, line 176: line 176 - instead of !Objects.nonNull. can you just use Objects.isNull?  should read better then

markpit: and noticed it in line 123

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -95,26 +99,98 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
...
+            beneficialOwnersCorporateSubmissionData.add(beneficialOwner);
+        }
+
+        return beneficialOwnersCorporateSubmissionData;
+    }
+
+    private List<TrustDataDto> getTrustData(OverseasEntitySubmissionDto submissionDto, List<String> trustIds, String noTrustsExceptionMessage) throws ServiceException {
+        List<TrustDataDto> trustsDataForBO = new ArrayList<>();
+        // Loop through each trustId(s) and lookup the information for that trust
+        if (Objects.nonNull(trustIds)) {
```
markpit, line 153: to reduce the cognitive complexity (see code smell) try ad ding a guard clause at top, replace line 53,

if (Objects.isNull(trustIds)) {
  return trustsDataForBO;
}

for (String trustId : trustIds) {
                List<TrustDataDto> trustData = new ArrayList<>();
etc....

spec/schema.json
mwestacott: I'm not suggesting changing the spec you've defined in any way, just one of the the examples. If it's an array of ids that come in, is that in a comma separated string? I might be missing the point here, in which case, let's have a quick call to dicuss.

mwestacott: Ok discussed this and understand the issue (when testing) so happy it stays as is.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingServiceTest.java
```diff
@@ -290,6 +311,14 @@ void testFilingGenerationWhenSuccessfulWithThreeBOCorporateTrustAndThreeBOIndivi
         OverseasEntitySubmissionDto overseasEntitySubmissionDto = Mocks.buildSubmissionDtoWithThreeBoCorporateTrustsAndThreeIndividualTrust();
         Optional<OverseasEntitySubmissionDto> submissionOpt = Optional.of(overseasEntitySubmissionDto);
         when(overseasEntitiesService.getOverseasEntitySubmission(OVERSEAS_ENTITY_ID)).thenReturn(submissionOpt);
+        when(objectMapper.writeValueAsString(any())).thenReturn(
+                "[{\"trust_id\":\"1\",\"trust_name\":\"Trust Name 1\",\"creation_date\":[2020,4,1],\"unable_to_obtain_all_trust_info\":null}]",
+                "[{\"trust_id\":\"2\",\"trust_name\":\"Trust Name 2\",\"creation_date\":[2020,4,1],\"unable_to_obtain_all_trust_info\":null}]",
+                "[{\"trust_id\":\"3\",\"trust_name\":\"Trust Name 3\",\"creation_date\":[2020,4,1],\"unable_to_obtain_all_trust_info\":null}]",
+                "[{\"trust_id\":\"1\",\"trust_name\":\"Trust Name 1\",\"creation_date\":[2020,4,1],\"unable_to_obtain_all_trust_info\":null}]",
```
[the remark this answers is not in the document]
jphillips2121: The trust ID's and names don't have to be unique when they're returning if it's for different BOs like in this instance. Here there are two BO's, one individual and one corporate which are both assigned to the same three trusts.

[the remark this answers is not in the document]
mwestacott: Right. Makes sense. Thanks for clarifying.

mwestacott (approved):

markpit (approved):

PR #80, opened 2022-08-08
ROE-1095 Add global exception handler

markpit:
https://companieshouse.atlassian.net/browse/ROE-1095

Add a generic exception handler to intercept uncaught exceptions and log them using Structured logger, so that the message gets a namespace for it to appear in Kibana logs

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/exception/GlobalExceptionHandler.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/exception/GlobalExceptionHandler.java
@@ -1,0 +1,21 @@
+package uk.gov.companieshouse.overseasentitiesapi.exception;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import uk.gov.companieshouse.overseasentitiesapi.utils.ApiLogger;
+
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/exception/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.HashMap;
+
+@ControllerAdvice
+public class GlobalExceptionHandler {
+
+    @ExceptionHandler(Exception.class)
+    public ResponseEntity<Object> handleException(Exception ex) {
+        HashMap<String, Object> logMap = new HashMap<>();
+        logMap.put("error", ex.getClass());
```
mwestacott, line 17: Should we also log the message separately here?

Like https://github.com/companieshouse/bankrupt-officer-search-api/blob/a9a6577cfea94635dad79d830fe7c7c3f672cfa5/src/main/java/uk/gov/companieshouse/bankruptofficersearch/api/exception/GlobalExceptionHandler.java#L22

Just need to ensure that output looks correct in Kibana. So maybe leave as is for now and see how the test in Toro1 pans out.

markpit: I could see the message logged from the message passed to the logger, but you might be right about kibana so i'll add it back in

mwestacott (approved):

PR #85, opened 2022-08-25
ROE-380 Add api validation for entityDTO

markpit:
https://companieshouse.atlassian.net/browse/ROE-380

Add Validation for the EntityDto submission data
Adding Entity name validation and unit tests
Feature flagged the validation call

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -513,6 +513,12 @@
         "start_date": {
           "type": "string",
           "format": "date"
+        },
+        "contact_full_name": {
+          "type": "string"
+        },
+        "contact_email": {
+          "type": "string"
         }
       },
       "required": [
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -435,6 +435,9 @@
           "items": {
             "$ref": "#/definitions/NatureOfControlType"
           }
+        },
+        "is_on_sanctions_list": {
+          "type": "boolean"
         }
       }
     },
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -218,6 +218,12 @@
           "items": {
             "$ref": "#/definitions/ManagingOfficerCorporate"
           }
+        },
+        "trusts": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/Trust"
+          }
         }
       },
       "required": [
@@ -339,6 +345,12 @@
         },
         "is_on_sanctions_list": {
           "type": "boolean"
+        },
+        "trust_ids": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
         }
       }
     },
@@ -396,6 +408,12 @@
         },
         "is_on_sanctions_list": {
           "type": "boolean"
+        },
+        "trust_ids": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
         }
       }
     },
@@ -438,6 +456,12 @@
         },
         "is_on_sanctions_list": {
           "type": "boolean"
+        },
+        "trust_ids": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
         }
       }
     },
@@ -530,6 +554,230 @@
         "is_on_register_in_country_formed_in",
         "start_date"
       ]
+    },
+    "Trust": {
+      "type": "object",
+      "properties": {
+        "trust_id": {
+          "type": "string"
+        },
+        "trust_name": {
+          "type": "string"
+        },
+        "creation_date": {
+          "type": "string",
+          "format": "date"
+        },
+        "unable_to_obtain_all_trust_info": {
+          "type": "boolean"
+        },
+        "INDIVIDUAL": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/TrustIndividual"
+          }
+        },
+        "CORPORATE": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/TrustCorporate"
+          }
+        },
+        "HISTORICAL_BO": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/HistoricalBeneficialOwner"
+          }
+        }
+      }
+    },
+    "TrustIndividual": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string"
+        },
+        "forename": {
+          "type": "string"
+        },
+        "other_forenames": {
+          "type": "string"
+        },
+        "surname": {
+          "type": "string"
+        },
+        "date_of_birth": {
+          "type": "string",
+          "format": "date"
+        },
+        "nationality": {
+          "type": "string"
+        },
+        "sa_address_line_1": {
+          "type": "string"
+        },
+        "sa_address_line_2": {
+          "type": "string"
+        },
+        "sa_address_care_of": {
+          "type": "string"
+        },
+        "sa_address_country": {
+          "type": "string"
+        },
+        "sa_address_locality": {
+          "type": "string"
+        },
+        "sa_address_po_box": {
+          "type": "string"
+        },
+        "sa_address_postal_code": {
+          "type": "string"
+        },
+        "sa_address_premises": {
+          "type": "string"
+        },
+        "sa_address_region": {
+          "type": "string"
+        },
+        "ura_address_line_1": {
+          "type": "string"
+        },
+        "ura_address_line_2": {
+          "type": "string"
+        },
+        "ura_address_care_of": {
+          "type": "string"
+        },
+        "ura_address_country": {
+          "type": "string"
+        },
+        "ura_address_locality": {
+          "type": "string"
+        },
+        "ura_address_po_box": {
+          "type": "string"
+        },
+        "ura_address_postal_code": {
+          "type": "string"
+        },
+        "ura_address_premises": {
+          "type": "string"
+        },
+        "ura_address_region": {
+          "type": "string"
+        },
+        "date_became_interested_person": {
+          "type": "string",
+          "format": "date"
+        }
+      }
+    },
+    "TrustCorporate": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string"
+        },
+        "name": {
+          "type": "string"
+        },
+        "sa_address_line_1": {
+          "type": "string"
+        },
+        "sa_address_line_2": {
+          "type": "string"
+        },
+        "sa_address_care_of": {
+          "type": "string"
+        },
+        "sa_address_country": {
+          "type": "string"
+        },
+        "sa_address_locality": {
+          "type": "string"
+        },
+        "sa_address_po_box": {
+          "type": "string"
+        },
+        "sa_address_postal_code": {
+          "type": "string"
+        },
+        "sa_address_premises": {
+          "type": "string"
+        },
+        "sa_address_region": {
+          "type": "string"
+        },
+        "ro_address_line_1": {
+          "type": "string"
+        },
+        "ro_address_line_2": {
+          "type": "string"
+        },
+        "ro_address_care_of": {
+          "type": "string"
+        },
+        "ro_address_country": {
+          "type": "string"
+        },
+        "ro_address_locality": {
+          "type": "string"
+        },
+        "ro_address_po_box": {
+          "type": "string"
+        },
+        "ro_address_postal_code": {
+          "type": "string"
+        },
+        "ro_address_premises": {
+          "type": "string"
+        },
+        "ro_address_region": {
+          "type": "string"
+        },
+        "identification_country_registration": {
+          "type": "string"
+        },
+        "identification_legal_authority": {
+          "type": "string"
+        },
+        "identification_legal_form": {
+          "type": "string"
+        },
+        "identification_place_registered": {
+          "type": "string"
+        },
+        "identification_registration_number": {
+          "type": "string"
+        },
+        "date_became_interested_person": {
+          "type": "string",
+          "format": "date"
+        }
+      }
+    },
+    "HistoricalBeneficialOwner": {
+      "type": "object",
+      "properties": {
+        "forename": {
+          "type": "string"
+        },
+        "other_forenames": {
+          "type": "string"
+        },
+        "surname": {
+          "type": "string"
+        },
+        "ceased_date": {
+          "type": "string",
+          "format": "date"
+        },
+        "notified_date": {
+          "type": "string",
+          "format": "date"
+        }
+      }
     },
     "Address": {
       "type": "object",
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -189,6 +189,9 @@
         "due_diligence" : {
           "$ref": "#/definitions/DueDiligence"
         },
+        "overseas_entity_due_diligence" : {
+          "$ref": "#/definitions/OverseasEntityDueDiligence"
+        },
         "beneficial_owners_statement": {
           "$ref": "#/definitions/BeneficialOwnersStatementType"
         },
@@ -309,7 +312,7 @@
         "name": {
           "type": "string"
         },
-        "address": {
+        "identity_address": {
           "$ref": "#/definitions/Address"
         },
         "email": {
@@ -328,6 +331,33 @@
           "type": "string"
         },
         "diligence": {
+          "type": "string"
+        }
+      }
+    },
+    "OverseasEntityDueDiligence": {
+      "type": "object",
+      "properties": {
+        "identity_date": {
+          "type": "string",
+          "format": "date"
+        },
+        "name": {
+          "type": "string"
+        },
+        "identity_address": {
+          "$ref": "#/definitions/Address"
+        },
+        "email": {
+          "type": "string"
+        },
+        "supervisory_name": {
+          "type": "string"
+        },
+        "aml_number": {
+          "type": "string"
+        },
+        "partner_name": {
           "type": "string"
         }
       }
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -186,6 +186,9 @@
         "entity": {
           "$ref": "#/definitions/Entity"
         },
+        "due_diligence" : {
+          "$ref": "#/definitions/DueDiligence"
+        },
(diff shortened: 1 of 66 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
```diff
@@ -0,0 +1,70 @@
...
+import uk.gov.companieshouse.overseasentitiesapi.validation.utils.StringValidators;
+import uk.gov.companieshouse.service.rest.err.Errors;
+
+@Component
+public class AddressDtoValidator {
+
+    public Errors validate(String parentAddressField, AddressDto addressDto, Errors errors, String loggingContext) {
+        validatePropertyNameNumber(parentAddressField, addressDto.getPropertyNameNumber(), errors, loggingContext);
+        validateLine1(parentAddressField, addressDto.getLine1(), errors, loggingContext);
+        validateLine2(parentAddressField, addressDto.getLine2(), errors, loggingContext);
```
mouhajer-ch, line 15: Before validating `line2` we need to make sure that the value is not null to avoid exception, and we should not use the `validateStringNotBlank` because is not a mandatory field

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/Country.java
```diff
@@ -0,0 +1,279 @@
...
+    ZAMBIA("Zambia"),
+    ZIMBABWE("Zimbabwe");
+
+    public final String countryName;
+
+    Country(String country) {
+        this.countryName = country;
+    }
+
+    public static boolean isValid(String countryIn, String location, Errors errs, String loggingContext) {
```
mouhajer-ch, line 268: We need to add a validation with all countries, including "United Kingdom","England","Wales","Scotland" and "Northern Ireland", and not only oversees countries.

The spec here https://github.com/companieshouse/overseas-entities-api/blob/7510143c3d23510807846c90b2d687ba8411e17a/spec/schema.json#L1169 has to change.
```
"anyOf": [
    {
        "enum": [
        "United Kingdom",
        "England",
        "Scotland",
        "Wales",
        "Northern Ireland"
        ]
    },
    {
    ...
    }
]
```

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
mattch1: Do we need this for the other optional fields

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
```diff
@@ -34,10 +37,16 @@
 public class OverseasEntitiesController {
 
     private final OverseasEntitiesService overseasEntitiesService;
+    private final OverseasEntitySubmissionDtoValidator overseasEntitySubmissionDtoValidator;
+
+    @Value("${FEATURE_FLAG_ENABLE_VALIDATION_25082022}")
```
mwestacott, line 42: Just curious, have we created a corresponding story on the backlog to remove this feature flag? If not, this needs to be done (at the point the feature flag is introduced).

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
```diff
@@ -47,17 +56,28 @@ public ResponseEntity<Object> createNewSubmission(
...
             @RequestHeader(value = ERIC_IDENTITY) String userId,
             HttpServletRequest request) {
-        String passthroughTokenHeader = request.getHeader(ApiSdkManager.getEricPassthroughTokenHeader());
 
         var logMap = new HashMap<String, Object>();
         logMap.put(TRANSACTION_ID_KEY, transaction.getId());
-        ApiLogger.infoContext(requestId, "Calling service to create Overseas Entity Submission", logMap);
 
         try {
+            if(isValidationEnabled) {
```
mwestacott, line 64: Formatting, but space needs adding between if and (.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
```diff
@@ -47,17 +56,28 @@ public ResponseEntity<Object> createNewSubmission(
...
+                var validationErrors = overseasEntitySubmissionDtoValidator.validate(overseasEntitySubmissionDto, new Errors(), requestId);
+
+                if (validationErrors.hasErrors()) {
+                    ApiLogger.infoContext(requestId, "Validation errors : " + validationErrors);
```
mwestacott, line 68: Could potentially log this as a warning (or even error) since, at the moment, any requests getting through with validation errors should have already been caught at the web layer and shouldn't have got this far.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/AddressDto.java
```diff
@@ -7,44 +7,55 @@
 
 public class AddressDto {
 
+    public static final String PROPERTY_NAME_NUMBER_FIELD = "property_name_number";
```
mwestacott, line 10: Would remove the _FIELD off the end of this constant name, it's not there on the others.

mwestacott: Oh, I see it is there in other DTOS. Maybe it needs adding on the others, or maybe these values are already used elsewhere (trusts?) and we should be careful about what we change... Need to be consistent with this naming, however. Requires some investigation.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
```diff
@@ -0,0 +1,34 @@
...
+            ApiLogger.infoContext(loggingContext, location + " date should be in the past");
+            return false;
+        }
+
+        return true;
+    }
+
+    public static boolean validateDateIsWithinLast3Months(LocalDate compareToDate, String location, Errors errs, String loggingContext) {
+        var localDateNow = LocalDate.now();
+        if (compareToDate.isBefore(localDateNow) && compareToDate.isAfter(localDateNow.minusMonths(3))) {
```
mwestacott, line 26: Implementation suggests that 'today' would cause 'false' to be returned but method name tells me that today should be accepted - 'validateDateIsWithinLast3Months'. Which is correct? Maybe method name needs tweaking...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.regex.Pattern;
+
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators.setErrorMsgToLocation;
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators.validateNotNull;
+
+public final class StringValidators {
+
+    private StringValidators() {}
+
+    public static boolean validateStringNotBlank(String toTest, String location, Errors errs, String loggingContext) {
```
mwestacott, line 15: Could drop 'String' from the method name here. It's in the StringValidators class and next max length method doesn't include 'String' in name.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.time.LocalDate;
+
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators.setErrorMsgToLocation;
+
+public class DateValidators {
+
+    private DateValidators() { }
+
+    public static boolean validateDateIsInPast(LocalDate compareToDate, String location, Errors errs, String loggingContext) {
```
mwestacott, line 14: All these method names (which return booleans) could just start 'is', as is convention. I think it would make code more readable.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
```diff
@@ -0,0 +1,70 @@
...
+        if (!matcher.matches()) {
+            setErrorMsgToLocation(errs, location, location + ValidationMessages.INVALID_CHARACTERS_ERROR_MESSAGE);
+            ApiLogger.infoContext(loggingContext, "Invalid characters for " + location);
+            return false;
+        }
+        return true;
+    }
+
+    public static boolean validateEmail(String email, String location, Errors errs, String loggingContext) {
+        // The email regex is taken from OWASP https://owasp.org/www-community/OWASP_Validation_Regex_Repository
```
mwestacott, line 44: Move this useful comment into a Javadoc comment on the method. When dealing with email addresses I like to get the 'address' part into the name as it's not the actual email we're validating. How about 'isValidEmailAddress' for the method name?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    public static boolean validateDateIsWithinLast3Months(LocalDate compareToDate, String location, Errors errs, String loggingContext) {
+        var localDateNow = LocalDate.now();
+        if (compareToDate.isBefore(localDateNow) && compareToDate.isAfter(localDateNow.minusMonths(3))) {
+            return true;
+        }
+
+        setErrorMsgToLocation(errs, location, location + ValidationMessages.DATE_NOT_WITHIN_PAST_3_MONTHS_ERROR_MESSAGE);
+        ApiLogger.infoContext(loggingContext, location + " date should be in the past");
```
mwestacott, line 31: Error message suggests that 'location' is the wrong variable name here. Should it be 'fieldName' or 'qualifiedFieldName' or something? Applies to all the other validation methods too.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/ValidationMessages.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.companieshouse.overseasentitiesapi.validation.utils;
+
+public class ValidationMessages {
+
+    private ValidationMessages() {}
+
+    public static final String NOT_NULL_ERROR_MESSAGE = " must not be null";
```
mwestacott, line 7: How about using %s in the error message content (like INVALID_EMAIL_ERROR_MESSAGE below) so as not to have to do the string concatenation, when used?

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidatorsTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        boolean isValidDate = DateValidators.validateDateIsInPast(FUTURE_DATE_TEST, LOCATION, errors, LOGGING_CONTEXT);
+
+        assertFalse(isValidDate);
+        assertEquals(1, errors.size());
+        assertTrue(errors.containsError(err));
+    }
+
+    @Test
+    @DisplayName("Validate date is within last 3 months successfully")
```
mwestacott, line 47: Add a boundary test for this with today as the input date. What should happen? Does it pass? (relates to a previous comment above)

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/Country.java
mwestacott: As agreed, this is not in the spec for the moment as it's not related to validation of the Overseas Entity information, so doesn't need to be implemented here yet either.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
mattch1: Updated this.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
mattch1: Log as an error added new exception as argument

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/AddressDto.java
mattch1: Updated names with FIELD suffix.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
mattch1: Not sure what to do about this.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
mattch1: Removed from this and other methods in this class.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
mattch1: Updated all booleans with "isValid"

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
mattch1: Added a javadoc.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
mattch1: All instances of location now say qualifiedFieldName

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/Country.java
mouhajer-ch: Nope, it is related to the validation of overseas entity. We need to differentiate between `incorporation_country` field and `principal-address` field

mouhajer-ch: I will do it

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
mouhajer-ch: Thanks Martin, I will create a story

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
mouhajer-ch: Hi Martin, any name suggestion?

mwestacott: So first question is, do we need this now? Web has a `checkDateIsWithinLast3Months()` function but seems to only be used by the 'due diligence' checks, so not for the 'tell us about the overseas entity screen'.

`checkDateIsWithinLast3Months()` in web does seem to include 'today' though, so name of this function might be fine (if it remains), just implementation that needs changing - start with a unit test to prove my theory that implementation is wrong...

mouhajer-ch: Thanks Martin, you are right,  we do not need it for the Entity was part of the Utils task, that probably should have been in a different task instead of a subtask. 
The date utils will be moved in a separate task. Anyway waiting your unit test to prove me wrong :D

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
mwestacott: Thanks. I don't think 'Valid' is needed in the names though. Compare 'isValidDateIsInPast' to 'isDateInPast'. I think second variation reads better when inline in the code. Minor though...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/ValidationMessages.java
mattch1: Updated this, significant unit test changes needed.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidators.java
mattch1: Date implementation changed to include today by checking date is before tomorrow (today plus a day) rather than before today.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DateValidatorsTest.java
mattch1: Added the boundary test with true expected.

spec/schema.json
```diff
@@ -1168,7 +1168,18 @@
       },
       "Country": {
         "type": "string",
-        "enum": [
+        "anyOf": [
```
mwestacott, line 1171: Not sure why this enum has come back in - I thought it wasn't required for just the entity validation. Anyway, no need for it to hold up merging and other work, so will approve.

mwestacott (approved):

PR #92, opened 2022-09-07
ROE-1252 Api spec for OE Due Diligence

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1252

Update api with validation for due diligence.

### DIFF
--- a/spec/schema.json
+++ b/spec/schema.json
@@ -438,7 +438,8 @@
       "OverseasEntityDueDiligence": {
         "required": [
           "name",
-          "email"
+          "email",
+          "identity_address"
         ],
         "type": "object",
         "properties": {
@@ -447,7 +448,6 @@
             "format": "date"
           },
           "name": {
-            "type": "string",
             "allOf" : [
               {
                 "$ref": "#/components/schemas/FieldPattern"
@@ -466,7 +466,6 @@
             "maxLength": 256
           },
           "supervisory_name": {
-            "type": "string",
             "allOf" : [
               {
                 "$ref": "#/components/schemas/FieldPattern"
@@ -481,7 +480,6 @@
             "maxLength": 256
           },
           "partner_name": {
-            "type": "string",
             "allOf" : [
               {
                 "$ref": "#/components/schemas/FieldPattern"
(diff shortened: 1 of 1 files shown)


spec/schema.json
```diff
@@ -436,29 +436,60 @@
         }
       },
       "OverseasEntityDueDiligence": {
+        "required": [
+          "name",
+          "email"
```
mouhajer-ch, line 441: `identity_address` is required

spec/schema.json
```diff
@@ -436,29 +436,60 @@
...
+              {
+                "maxLength": 256
+              }
+            ]
           },
           "identity_address": {
             "$ref": "#/components/schemas/Address"
           },
           "email": {
-            "type": "string"
```
mouhajer-ch, line 452: Probably better if we create a component schema for email and use `$ref`

mouhajer-ch (commented):
We need to update the `Address` schema to consider that the  `county` and `postcode` are required and the `property_name_number` is not in due diligence models
```
"Address": {
  "type": "object",
  ...,
  "required": [
    "line_1",
    "town",
    "country"
  ],
  "oneOf": [
    {
      "required": [
        "property_name_number"
      ]
    },
    {
      "required": [
        "county",
        "postcode"
      ]
    }
  ]
},
```

spec/schema.json
mouhajer-ch: Lets do it when we have consistency between emails, already discussed with Francis to check why we have got some fields with length 256 and other 250 when the table can accept 256.

mouhajer-ch:
> We need to update the `Address` schema to consider that the `county` and `postcode` are required and the `property_name_number` is not in due diligence models
> 
> ```
> "Address": {
>   "type": "object",
>   ...,
>   "required": [
>     "line_1",
>     "town",
>     "country"
>   ],
>   "oneOf": [
>     {
>       "required": [
>         "property_name_number"
>       ]
>     },
>     {
>       "required": [
>         "county",
>         "postcode"
>       ]
>     }
>   ]
> },
> ```

Fixed in the following PR https://github.com/companieshouse/overseas-entities-api/pull/93

markpit (approved):

PR #100, opened 2022-09-15
ROE-1181 Avoid unexpected test failure with test dates

mattch1:
Dates in unit tests can come back to haunt to added a an ensured past date in the before each code.


mwestacott (approved):
Assume this failed sporadically before? Anyway, looks sensible.

mattch1:
It wasn't failing but Sam's tests were and so I suggested adding a date based on now, to before each. This fixed his problem. I then checked whether I had done the same (both working on the due diligence pages) and I had not, so decided it would be wise to update it.

PR #99, opened 2022-09-15
Roe 382 individual bo validation

mattch1:
https://companieshouse.atlassian.net/browse/ROE-382

Spec update

Also updated the handling of the country list as this will require both overseas and uk countries.

The nature of control fields all need to be tested together so added an umbrella nature of control variable as what is in the dto is too fine grained.

PR #102, opened 2022-09-26
ROE-382 Swagger spec changes for NOC mandatory fields

mwestacott:


spec/schema.json
```diff
@@ -550,6 +550,24 @@
...
+        "anyOf": [
+          {
+            "required": [
+              "beneficial_owner_nature_of_control_types"
+            ]
+          },
+          {
+            "required": [
+              "trustees_nature_of_control_types",
+              "trust_ids"
```
mwestacott, line 562: Just wonder if 'trust_ids' should feature here (I took this from your previous suggestion @mouhajer-ch)?

Due to the 'anyOf' clause above, is it saying that the client could provide only some trust ids and no NOC type for the Individual BO, because that doesn't sound right?

markpit: I 'think' by having the trustees and trust_ids in the same list it is joining them with an 'and'. So if they provide a trustee noc type than it must also have trust_ids values on the BO individual. Which sounds ok.

mouhajer-ch: It looks good to me, as I have suggested here: https://github.com/companieshouse/overseas-entities-api/pull/99#discussion_r978441233

mouhajer-ch (approved):

spec/schema.json
mwestacott: Thanks both. Will merge.

PR #105, opened 2022-09-29
ROE-1311 No service address when same as flag is true

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1311

Validation has to ensure that flag controlled information is not unnecessarily passed into the api when it is not expected, in this case when the same address flag is true for entity and all B.Os, the service address is empty.

From looking at the web an empty Service address and never a null one is passed into the api.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
@@ -34,16 +34,16 @@ public class AddressDtoValidator {
     }
 
     public Errors validateOtherAddressIsNotSupplied(String parentAddressField, AddressDto addressDto, Errors errors, String loggingContext) {
-        StringValidators.checkisEmpty(addressDto.getPropertyNameNumber(), getQualifiedFieldName(parentAddressField, AddressDto.PROPERTY_NAME_NUMBER_FIELD), errors, loggingContext);
-        StringValidators.checkisEmpty(addressDto.getLine1(), getQualifiedFieldName(parentAddressField, AddressDto.LINE_1_FIELD), errors, loggingContext);
-        StringValidators.checkisEmpty(addressDto.getLine2(), getQualifiedFieldName(parentAddressField, AddressDto.LINE_2_FIELD), errors, loggingContext);
-        StringValidators.checkisEmpty(addressDto.getTown(), getQualifiedFieldName(parentAddressField, AddressDto.TOWN_FIELD), errors, loggingContext);
-        StringValidators.checkisEmpty(addressDto.getCounty(), getQualifiedFieldName(parentAddressField, AddressDto.COUNTY_FIELD), errors, loggingContext);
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
```diff
@@ -75,4 +75,11 @@ private static boolean isNotEmpty(String toTest, String qualifiedFieldName, Erro
         }
         return true;
     }
+
+    public static void checkisEmpty(String toTest, String qualifiedFieldName, Errors errs, String loggingContext) {
```
markpit, line 79: the 'checkisEmpty' name should have a capital 'I' for 'is'

mattch1: Corrected.

mouhajer-ch (dismissed):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
```diff
@@ -33,6 +33,20 @@ public Errors validate(String parentAddressField, AddressDto addressDto, List<St
         return errors;
     }
 
+    public Errors validateOtherAddressIsNotSupplied(String parentAddressField, AddressDto addressDto, Errors errors, String loggingContext) {
+        StringValidators.checkIsEmpty(addressDto.getPropertyNameNumber(), getQualifiedFieldName(parentAddressField, AddressDto.PROPERTY_NAME_NUMBER_FIELD), errors, loggingContext);
```
mwestacott, line 37: Been chatting to @markpit , we think this block of checks should be wrapped in 'not null' check, i.e. only do them if the 'addressDto' is non null. If it is null, there is no error to report as that means the alternative address has not been supplied.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
```diff
@@ -75,4 +75,11 @@ private static boolean isNotEmpty(String toTest, String qualifiedFieldName, Erro
         }
         return true;
     }
+
+    public static void checkIsEmpty(String toTest, String qualifiedFieldName, Errors errs, String loggingContext) {
+        if (toTest != null && !toTest.trim().isEmpty()) {
```
mwestacott, line 80: Use Object.nonNull() for the first check.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
```diff
@@ -75,4 +75,11 @@ private static boolean isNotEmpty(String toTest, String qualifiedFieldName, Erro
...
+            setErrorMsgToLocation(errs, qualifiedFieldName, ValidationMessages.SHOULD_NOT_BE_POPULATED_ERROR_MESSAGE.replace("%s", qualifiedFieldName));
```
mwestacott, line 81: I would suggest using the String 'format' method here, rather than 'replace'. Examples already elsewhere.

mwestacott (approved):

PR #109, opened 2022-10-03
ROE-1339 Should fail validation if both due diligence options are sup…

mattch1:
…plied.

https://companieshouse.atlassian.net/browse/ROE-1339

Introduced dedicated code to validate both due diligence dtos because the web never sends a null value only an empty one so it is the content that needs to be checked.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
@@ -1,8 +1,10 @@
 package uk.gov.companieshouse.overseasentitiesapi.model.dto;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.commons.lang.StringUtils;
 
 import java.time.LocalDate;
+import java.util.Objects;
 
 public class DueDiligenceDto {
 
@@ -113,5 +115,17 @@ public class DueDiligenceDto {
 
     public void setDiligence(String diligence) {
         this.diligence = diligence;
+    }
+
+    public boolean isEmpty() {
+        return Objects.isNull(identityDate) &&
+               StringUtils.isBlank(name) &&
+               Objects.isNull(address) &&
+               StringUtils.isBlank(email) &&
+               StringUtils.isBlank(supervisoryName) &&
+               StringUtils.isBlank(amlNumber) &&
+               StringUtils.isBlank(agentCode) &&
+               StringUtils.isBlank(partnerName) &&
+               StringUtils.isBlank(diligence);
     }
 }
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
@@ -2,6 +2,8 @@ package uk.gov.companieshouse.overseasentitiesapi.validation;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.overseasentitiesapi.model.dto.DueDiligenceDto;
+import uk.gov.companieshouse.overseasentitiesapi.model.dto.OverseasEntityDueDiligenceDto;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.OverseasEntitySubmissionDto;
 import uk.gov.companieshouse.overseasentitiesapi.validation.utils.DueDiligenceOptionValidators;
 import uk.gov.companieshouse.service.rest.err.Errors;
@@ -44,15 +46,14 @@ public class OverseasEntitySubmissionDtoValidator {
         entityDtoValidator.validate(overseasEntitySubmissionDto.getEntity(), errors, loggingContext);
         presenterDtoValidator.validate(overseasEntitySubmissionDto.getPresenter(), errors, loggingContext);
 
-        if (DueDiligenceOptionValidators.bothDueDiligenceOptionsNotSupplied(
-                overseasEntitySubmissionDto.getDueDiligence(), overseasEntitySubmissionDto.getOverseasEntityDueDiligence(),
-                errors,
-                loggingContext)) {
+        DueDiligenceDto dueDiligenceDto = overseasEntitySubmissionDto.getDueDiligence();
+        OverseasEntityDueDiligenceDto overseasEntityDueDiligenceDto = overseasEntitySubmissionDto.getOverseasEntityDueDiligence();
+        if (DueDiligenceOptionValidators.bothDueDiligenceOptionsNotSupplied(dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
 
-            if (Objects.nonNull(overseasEntitySubmissionDto.getOverseasEntityDueDiligence())) {
+            if (Objects.nonNull(overseasEntityDueDiligenceDto) && !overseasEntityDueDiligenceDto.isEmpty()) {
                 overseasEntityDueDiligenceValidator.validate(overseasEntitySubmissionDto.getOverseasEntityDueDiligence(), errors, loggingContext);
             }
-            if (Objects.nonNull(overseasEntitySubmissionDto.getDueDiligence())) {
+            if (Objects.nonNull(dueDiligenceDto) && !dueDiligenceDto.isEmpty()) {
                 dueDiligenceValidator.validate(overseasEntitySubmissionDto.getDueDiligence(), errors, loggingContext);
             }
         }
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
@@ -49,7 +49,7 @@ public class OverseasEntitySubmissionDtoValidator {
 
         var dueDiligenceDto = overseasEntitySubmissionDto.getDueDiligence();
         var overseasEntityDueDiligenceDto = overseasEntitySubmissionDto.getOverseasEntityDueDiligence();
-        if (dueDiligenceDataBlockValidator.onlyOneDueDiligencePresent(dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
+        if (dueDiligenceDataBlockValidator.onlyOneBlockPresent(dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
 
             if (Objects.nonNull(overseasEntityDueDiligenceDto) && !overseasEntityDueDiligenceDto.isEmpty()) {
                 overseasEntityDueDiligenceValidator.validate(overseasEntitySubmissionDto.getOverseasEntityDueDiligence(), errors, loggingContext);
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
@@ -5,7 +5,7 @@ import org.springframework.stereotype.Component;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.DueDiligenceDto;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.OverseasEntityDueDiligenceDto;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.OverseasEntitySubmissionDto;
-import uk.gov.companieshouse.overseasentitiesapi.validation.utils.DueDiligenceOptionValidators;
+import uk.gov.companieshouse.overseasentitiesapi.validation.utils.DueDiligenceDataBlockValidators;
 import uk.gov.companieshouse.service.rest.err.Errors;
 
 import java.util.Objects;
@@ -46,14 +46,13 @@ public class OverseasEntitySubmissionDtoValidator {
         entityDtoValidator.validate(overseasEntitySubmissionDto.getEntity(), errors, loggingContext);
         presenterDtoValidator.validate(overseasEntitySubmissionDto.getPresenter(), errors, loggingContext);
 
-        DueDiligenceDto dueDiligenceDto = overseasEntitySubmissionDto.getDueDiligence();
-        OverseasEntityDueDiligenceDto overseasEntityDueDiligenceDto = overseasEntitySubmissionDto.getOverseasEntityDueDiligence();
-        if (DueDiligenceOptionValidators.onlyOneDueDiligencePresent(dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
+        var dueDiligenceDto = overseasEntitySubmissionDto.getDueDiligence();
+        var overseasEntityDueDiligenceDto = overseasEntitySubmissionDto.getOverseasEntityDueDiligence();
+        if (DueDiligenceDataBlockValidators.onlyOneDueDiligencePresent(dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
 
             if (Objects.nonNull(overseasEntityDueDiligenceDto) && !overseasEntityDueDiligenceDto.isEmpty()) {
                 overseasEntityDueDiligenceValidator.validate(overseasEntitySubmissionDto.getOverseasEntityDueDiligence(), errors, loggingContext);
-            }
-            if (Objects.nonNull(dueDiligenceDto) && !dueDiligenceDto.isEmpty()) {
+            } else if (Objects.nonNull(dueDiligenceDto) && !dueDiligenceDto.isEmpty()) {
                 dueDiligenceValidator.validate(overseasEntitySubmissionDto.getDueDiligence(), errors, loggingContext);
             }
         }
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
@@ -3,7 +3,9 @@ package uk.gov.companieshouse.overseasentitiesapi.validation;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.OverseasEntitySubmissionDto;
+import uk.gov.companieshouse.overseasentitiesapi.validation.utils.DueDiligenceOptionValidators;
 import uk.gov.companieshouse.service.rest.err.Errors;
+
 import java.util.Objects;
 
 @Component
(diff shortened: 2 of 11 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -41,11 +45,17 @@ public OverseasEntitySubmissionDtoValidator(EntityDtoValidator entityDtoValidato
...
         presenterDtoValidator.validate(overseasEntitySubmissionDto.getPresenter(), errors, loggingContext);
-        if(Objects.nonNull(overseasEntitySubmissionDto.getOverseasEntityDueDiligence())) {
-            overseasEntityDueDiligenceValidator.validate(overseasEntitySubmissionDto.getOverseasEntityDueDiligence(), errors, loggingContext);
-        }
-        if(Objects.nonNull(overseasEntitySubmissionDto.getDueDiligence())) {
-            dueDiligenceValidator.validate(overseasEntitySubmissionDto.getDueDiligence(), errors, loggingContext);
+
+        DueDiligenceDto dueDiligenceDto = overseasEntitySubmissionDto.getDueDiligence();
+        OverseasEntityDueDiligenceDto overseasEntityDueDiligenceDto = overseasEntitySubmissionDto.getOverseasEntityDueDiligence();
+        if (DueDiligenceOptionValidators.bothDueDiligenceOptionsNotSupplied(dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
```
mouhajer-ch, line 51: We need to check that only one of the DueDiligence is present we can not have both and we can not allow both empty, last one has been covered.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceOptionValidators.java
```diff
@@ -0,0 +1,30 @@
...
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.ValidationMessages.SHOULD_NOT_BOTH_BE_POPULATED_ERROR_MESSAGE;
+
+public class DueDiligenceOptionValidators {
+
+     public static boolean bothDueDiligenceOptionsNotSupplied(DueDiligenceDto dueDiligenceDto,
+                                                              OverseasEntityDueDiligenceDto overseasEntityDueDiligenceDto,
+                                                              Errors errs,
+                                                              String loggingContext) {
+
+         if ((Objects.nonNull(dueDiligenceDto) && Objects.nonNull(overseasEntityDueDiligenceDto)) &&
```
mouhajer-ch, line 21: I do not understand these conditions, probably you mean both not null and not empty

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceOptionValidators.java
```diff
@@ -0,0 +1,30 @@
...
+import uk.gov.companieshouse.service.rest.err.Errors;
+
+import java.util.Objects;
+
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators.setErrorMsgToLocation;
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.ValidationMessages.SHOULD_NOT_BOTH_BE_POPULATED_ERROR_MESSAGE;
+
+public class DueDiligenceOptionValidators {
+
+     public static boolean bothDueDiligenceOptionsNotSupplied(DueDiligenceDto dueDiligenceDto,
```
mouhajer-ch, line 16: Probably we should change the name, what do you think :)

markpit:
there's 3 code smells to be looked at

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
```diff
@@ -114,4 +116,16 @@ public String getDiligence() {
     public void setDiligence(String diligence) {
         this.diligence = diligence;
     }
+
+    public boolean isEmpty() {
```
mwestacott, line 120: Not sure we need this isEmpty check - do we need to be this clever? Client should only submit one due diligence block. Checking that both due diligence blocks are present should be sufficient to fix the defect, as it's documented.

Would be good to get some other views on this though...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
```diff
@@ -114,4 +116,16 @@ public String getDiligence() {
...
+        return Objects.isNull(identityDate) &&
+               StringUtils.isBlank(name) &&
+               Objects.isNull(address) &&
```
mwestacott, line 123: If we do go down this route, this line would/should probably call an isEmpty() method on the address object, which would need implementing.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceOptionValidators.java
```diff
@@ -0,0 +1,30 @@
...
+import uk.gov.companieshouse.overseasentitiesapi.model.dto.OverseasEntitySubmissionDto;
+import uk.gov.companieshouse.overseasentitiesapi.utils.ApiLogger;
+import uk.gov.companieshouse.service.rest.err.Errors;
+
+import java.util.Objects;
+
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators.setErrorMsgToLocation;
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.ValidationMessages.SHOULD_NOT_BOTH_BE_POPULATED_ERROR_MESSAGE;
+
+public class DueDiligenceOptionValidators {
```
mwestacott, line 14: Rename to DueDiligenceDataBlockValidator, perhaps?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott, line 51: See comment below with suggested class rename, but if adopted, this method name could become 'bothBlocksSupplied()' to improve readability of the code.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -41,11 +45,17 @@ public OverseasEntitySubmissionDtoValidator(EntityDtoValidator entityDtoValidato
...
+
+            if (Objects.nonNull(overseasEntityDueDiligenceDto) && !overseasEntityDueDiligenceDto.isEmpty()) {
+                overseasEntityDueDiligenceValidator.validate(overseasEntitySubmissionDto.getOverseasEntityDueDiligence(), errors, loggingContext);
+            }
```
mwestacott, line 55: Suggest doing an 'else' at this point, rather than a new 'if' block. We know/expect that if one block isn't supplied, the other must be.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceOptionValidators.java
mwestacott: Yes, have added a suggestion for the method in the class above.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceOptionValidators.java
```diff
@@ -0,0 +1,30 @@
...
+     public static boolean bothDueDiligenceOptionsNotSupplied(DueDiligenceDto dueDiligenceDto,
+                                                              OverseasEntityDueDiligenceDto overseasEntityDueDiligenceDto,
+                                                              Errors errs,
+                                                              String loggingContext) {
+
+         if ((Objects.nonNull(dueDiligenceDto) && Objects.nonNull(overseasEntityDueDiligenceDto)) &&
+                 !(dueDiligenceDto.isEmpty() || overseasEntityDueDiligenceDto.isEmpty())) {
+             String qualifiedFieldNames = OverseasEntitySubmissionDto.DUE_DILIGENCE_FIELD + " and " + OverseasEntitySubmissionDto.OVERSEAS_ENTITY_DUE_DILIGENCE;
+             setErrorMsgToLocation(errs, qualifiedFieldNames, String.format(SHOULD_NOT_BOTH_BE_POPULATED_ERROR_MESSAGE, qualifiedFieldNames));
+             ApiLogger.infoContext(loggingContext, "Both the " + qualifiedFieldNames + " Fields should not be populated");
```
mwestacott: So log message needs to change here I think. Perhaps call them 'blocks' rather than fields?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
mattch1: See my original comment above, the web sends empty objects not null ones, so it would be difficult (or impossible) to implement this without that method.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
mattch1: A task for another ticket eprhaps.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mattch1: Updated this, see what you think.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceOptionValidators.java
mattch1: Updated the messages.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
mwestacott: Maybe. What does the web send down here, out of interest, when the due diligence block is empty/unpopulated? A null or an address block with all empty fields?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
mwestacott: 👍 missed that. Makes sense now.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -41,11 +45,16 @@ public OverseasEntitySubmissionDtoValidator(EntityDtoValidator entityDtoValidato
...
         presenterDtoValidator.validate(overseasEntitySubmissionDto.getPresenter(), errors, loggingContext);
-        if(Objects.nonNull(overseasEntitySubmissionDto.getOverseasEntityDueDiligence())) {
-            overseasEntityDueDiligenceValidator.validate(overseasEntitySubmissionDto.getOverseasEntityDueDiligence(), errors, loggingContext);
-        }
-        if(Objects.nonNull(overseasEntitySubmissionDto.getDueDiligence())) {
-            dueDiligenceValidator.validate(overseasEntitySubmissionDto.getDueDiligence(), errors, loggingContext);
+
+        var dueDiligenceDto = overseasEntitySubmissionDto.getDueDiligence();
+        var overseasEntityDueDiligenceDto = overseasEntitySubmissionDto.getOverseasEntityDueDiligence();
+        if (DueDiligenceDataBlockValidators.onlyOneDueDiligencePresent(dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
```
mwestacott, line 51: I would get 'Block' into the method name here - 'onlyOneBlockPresent' perhaps. Up to you though, already reads better than before.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceDataBlockValidators.java
```diff
@@ -0,0 +1,60 @@
...
+                  dueDiligenceDto.isEmpty() && overseasEntityDueDiligenceDto.isEmpty()) {
+
+              logValidationErrorMessage(errors, loggingContext, SHOULD_NOT_BOTH_BE_ABSENT_ERROR_MESSAGE);
+              return true;
+          }
+          return false;
+    }
+
+    private static void logValidationErrorMessage(Errors errors, String loggingContext, String errorMessage) {
+        String qualifiedFieldNames = OverseasEntitySubmissionDto.DUE_DILIGENCE_FIELD + " and " + OverseasEntitySubmissionDto.OVERSEAS_ENTITY_DUE_DILIGENCE;
```
mwestacott, line 56: This could be a static final class string constant as it is always the same value.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceDataBlockValidators.java
```diff
@@ -0,0 +1,60 @@
...
+import uk.gov.companieshouse.overseasentitiesapi.utils.ApiLogger;
+import uk.gov.companieshouse.service.rest.err.Errors;
+
+import java.util.Objects;
+
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators.setErrorMsgToLocation;
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.ValidationMessages.SHOULD_NOT_BOTH_BE_ABSENT_ERROR_MESSAGE;
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.ValidationMessages.SHOULD_NOT_BOTH_BE_PRESENT_ERROR_MESSAGE;
+
+public class DueDiligenceDataBlockValidators {
```
mwestacott, line 15: Would suggest that class name is singular not plural.

You also need a unit test for this class - the logic is not trivial. I'm having a little trouble understanding/reading it so some well named unit tests documenting the allowed and disallowed scenarios would be really helpful.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott: New comment added so resolving this one. Suggest a minor change though.

mattch1:
I think this needs some discussion, I am not sure what the right strategy is here.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mattch1: Renamed

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -41,11 +45,16 @@ public OverseasEntitySubmissionDtoValidator(EntityDtoValidator entityDtoValidato
...
+
+            if (Objects.nonNull(overseasEntityDueDiligenceDto) && !overseasEntityDueDiligenceDto.isEmpty()) {
+                overseasEntityDueDiligenceValidator.validate(overseasEntitySubmissionDto.getOverseasEntityDueDiligence(), errors, loggingContext);
+            } else if (Objects.nonNull(dueDiligenceDto) && !dueDiligenceDto.isEmpty()) {
```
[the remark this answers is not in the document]
mattch1: Just says "else" now.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceDataBlockValidators.java
mattch1: Extracted static variable.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceDataBlockValidators.java
mattch1: Singular

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceDataBlockValidators.java
```diff
@@ -0,0 +1,60 @@
...
+
+     private DueDiligenceDataBlockValidators(){}
+
+     public static boolean onlyOneDueDiligencePresent(DueDiligenceDto dueDiligenceDto,
```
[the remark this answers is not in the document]
mattch1: Refactored this to be non static and added 10 new unit tests.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
mattch1: 10 new unit tests

mwestacott: Thanks for adding those tests.

What does the web send though please, when block is unpopulated? And could you name the specific test(s) that check this scenario?

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidatorTest.java
```diff
@@ -73,12 +77,13 @@ class OverseasEntitySubmissionDtoValidatorTest {
...
-    void testOverseasEntityDueDiligenceValidator() {
+    void testOverseasEntitySubmissionValidator() {
+        when(dueDiligenceDataBlockValidator.onlyOneBlockPresent(any(), any(), any(), any())).thenReturn(true);
         buildOverseasEntitySubmissionDto();
-        overseasEntitySubmissionDtoValidator.validate(overseasEntitySubmissionDto, new Errors(), CONTEXT);
+        overseasEntitySubmissionDtoValidator.validate(overseasEntitySubmissionDto, new Errors(), LOGGING_CONTEXT);
         verify(entityDtoValidator, times(1)).validate(eq(entityDto),any(),any());
         verify(presenterDtoValidator, times(1)).validate(eq(presenterDto),any(),any());
-        verify(overseasEntityDueDiligenceValidator, times(1)).validate(eq(overseasEntityDueDiligenceDto),any(),any());
+        verify(overseasEntityDueDiligenceValidator, times(0)).validate(eq(overseasEntityDueDiligenceDto),any(),any());
```
markpit, line 86: i think we'll also need to test that the overseasEntityDueDiligenceValidator gets called too. Maybe have another @Test method, identical to the testOverseasEntitySubmissionValidator, but have it set the overseasEntityDueDiligenceValidator instead of the dueDiligenceValidator, so we can be sure that each validator gets called when each of the data blocks are present.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/DueDiligenceDto.java
mattch1: In the web we have:     
// Empty OverseasEntityDueDiligence object
    setApplicationData(req.session, {}, OverseasEntityDueDiligenceKey);

Same for the other block. 

The tests in DueDiligenceDataBlockValidatorTest deal with the different combinations and permutations of this scenario.

mattch1: Where you see "new DueDiligenceDto()" in the argument list that is an empty object.

mwestacott (approved):

markpit (approved):

PR #123, opened 2022-10-13
ROE-1227 Sanitise input data used in validation logging

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1227

Unable to find any input data logged other than two instances of country, these are unlikely to require truncation, but have sanitised and tested them.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
@@ -1,5 +1,6 @@
 package uk.gov.companieshouse.overseasentitiesapi.validation;
 
+import org.owasp.encoder.Encode;
 import org.springframework.stereotype.Component;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.AddressDto;
 import uk.gov.companieshouse.overseasentitiesapi.utils.ApiLogger;
@@ -93,7 +94,8 @@ public class AddressDtoValidator {
         if (countryNotBlank) {
             boolean isOnList = allowedCountries.contains(country);
             if (!isOnList) {
-                var validationMessage = String.format(ValidationMessages.COUNTRY_NOT_ON_LIST_ERROR_MESSAGE, country);
+                String sanitisedCountryName = Encode.forJava(country);
+                var validationMessage = String.format(ValidationMessages.COUNTRY_NOT_ON_LIST_ERROR_MESSAGE, sanitisedCountryName);
                 setErrorMsgToLocation(errors, qualifiedFieldName, validationMessage);
                 ApiLogger.infoContext(loggingContext, validationMessage);
             }
--- a/src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidatorTest.java
+++ b/src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidatorTest.java
@@ -4,9 +4,6 @@ import org.apache.commons.lang.StringUtils;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.NullSource;
-import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.junit.jupiter.MockitoExtension;
 import uk.gov.companieshouse.overseasentitiesapi.mocks.AddressMock;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.AddressDto;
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
```diff
@@ -93,7 +94,8 @@ private void validateCountry(String parentAddressField, String country, List<Str
         if (countryNotBlank) {
             boolean isOnList = allowedCountries.contains(country);
             if (!isOnList) {
-                var validationMessage = String.format(ValidationMessages.COUNTRY_NOT_ON_LIST_ERROR_MESSAGE, country);
+                String sanitisedCountryName = Encode.forJava(country);
```
mwestacott, line 97: We also need to truncate this field value. Here and in the validator below.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidatorTest.java
```diff
@@ -120,13 +117,37 @@ void testErrorReportedWhenServiceAddressIsNull() {
         String validationMessage = String.format(ValidationMessages.NOT_NULL_ERROR_MESSAGE, EntityDto.SERVICE_ADDRESS_FIELD);
         assertError(parentField, "", validationMessage, errors);
     }
+    @Test
```
mwestacott, line 120: I don't think you've added this, have you? It's already in the test class so presume it's GitHub getting confused...

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidatorTest.java
```diff
@@ -120,13 +117,37 @@ void testErrorReportedWhenServiceAddressIsNull() {
...
+        Errors errors = addressDtoValidator.validate(parentField, addressDto, CountryLists.getAllCountries(), new Errors(), LOGGING_CONTEXT);
+
+        String validationMessage = String.format(ValidationMessages.COUNTRY_NOT_ON_LIST_ERROR_MESSAGE, input);
+        assertError(parentField, AddressDto.COUNTRY_FIELD, validationMessage, errors);
+    }
 
-    void testErrorReportedWhenFictionalCountryIsNotTheListOfAllCountries(String input) {
-        addressDto.setCountry("Utopia");
+    @Test
+    void testErrorReportedWithoutUnSanitisedStringWhenUnsanitizedCountryIsInput() {
```
mwestacott, line 132: I'm a bit confused by these two test methods. They both seem to have the same set-up fixture and both call 'validate', so won't the outcome be the same in both methods? I think perhaps the naming of them could be improved - happy to have a call if you wish.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidatorTest.java
mattch1: No it had to be added. The parameterised test annotations had been removed and there were unused imports that had to be removed as well.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
mattch1: I'm not sure that we need to truncate this. I cannot think of any country names over 50 characters long.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidatorTest.java
mattch1: Let me try to make the names clearer and if they are not after that then yes we should have a call.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
mwestacott: The whole point of this work is if a malicious user is submitting bad data directly to our API to cause CH problems - in this case it would be log forging, filling the logs up or potentially causing some kind of service crash/error.

Because in the submited JSON body a very large country name could be submitted (not one that features in the enum) and is output, we need to do the truncation, just as we did for Confirmation Statement.

Hope that's clear but give me a call if not.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidatorTest.java
mwestacott: Ah, now I see.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidatorTest.java
mwestacott: Ok, now I follow what you're doing. In theory, you could just have one test method with both/more assertions (as fixture is the same in each) but no harm with it being more verbose.

Thanks for changing those method names.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/AddressDtoValidator.java
```diff
@@ -93,7 +95,11 @@ private void validateCountry(String parentAddressField, String country, List<Str
...
```
markpit, line 98: the sanitise and length truncation are done for a couple of fields so rather than duplicate the code to do that, I would make a new Sanitiser class under utils, have a method 'makeSafeForLogging'? that takes a string, Encodes it and truncates it if > 50 and returns the 'safe' string. That way we can keep the 'make safe' logic in one place and can unit test it. Have it as a @Component that can be injected where it is needed.

mattch1: Refactored and tested this.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/utils/DataSanitisationTest.java
```diff
@@ -0,0 +1,15 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class DataSanitisationTest {
+
+    @Test
+    void testMakeStringSafeForLogging() {
+        DataSanitisation dataSanitisation = new DataSanitisation();
+        String sanitisedInput = dataSanitisation.makeStringSafeForLogging("abc\t\nabc");
```
markpit, line 12: make the string to be sanitised > 50 chars as well to test that it truncates it

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/BeneficialOwnerCorporateValidatorTest.java
```diff
@@ -34,7 +35,7 @@ class BeneficialOwnerCorporateValidatorTest {
 
     @BeforeEach
     public void init() {
-        beneficialOwnerCorporateValidator = new BeneficialOwnerCorporateValidator(new AddressDtoValidator());
+        beneficialOwnerCorporateValidator = new BeneficialOwnerCorporateValidator(new AddressDtoValidator(new DataSanitisation()));
```
mwestacott, line 38: Note that because you're creating an actual instance of the new sanitiser class and not using a mock this 'unit test' is now essentially testing the behaviour of the BeneficialOwnerCorporateValidator code AND the DataSanitisation code.

Would be better to mock it, I think, but up to you.

mattch1: Added inject mocks to addresses and mocked data sanitisation where needed.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/utils/DataSanitisationTest.java
mattch1: Added another test for this.

mwestacott (approved):

markpit (approved):

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/BeneficialOwnerCorporateValidatorTest.java
mwestacott: Thanks @mattch1 .

PR #125, opened 2022-10-17
Roe 1431 missing page data null checks

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1304

Added null checks to the first two fields. The tests just verify that other validation checks get called when these fields are null, as mocks cannot be injected two layers deep.

Added the who is registering field to the dto model only. This will need to be passed form the web.
Added extra condition gate to check that incorrect data has not been submitted against the who is registering field.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
@@ -23,7 +23,6 @@ public class OverseasEntitySubmissionDto {
     public static final String MANAGING_OFFICERS_INDIVIDUAL_FIELD = "managing_officers_individual";
     public static final String MANAGING_OFFICERS_CORPORATE_FIELD = "managing_officers_corporate";
     public static final String TRUST_DATA = "trusts";
-    public static final String WHO_IS_REGISTERING = "who_is_registering";
 
     @JsonProperty(PRESENTER_FIELD)
     private PresenterDto presenter;
@@ -58,9 +57,6 @@ public class OverseasEntitySubmissionDto {
     @JsonInclude(NON_NULL)
     @JsonProperty(TRUST_DATA)
     private List<TrustDataDto> trusts;
-
-    @JsonProperty(WHO_IS_REGISTERING)
-    private String whoIsRegistering;
 
     @JsonProperty("links")
     private Map<String, String> links;
@@ -159,13 +155,5 @@ public class OverseasEntitySubmissionDto {
 
     public void setLinks(Map<String, String> links) {
         this.links = links;
-    }
-
-    public String getWhoIsRegistering() {
-        return whoIsRegistering;
-    }
-
-    public void setWhoIsRegistering(String whoIsRegistering) {
-        this.whoIsRegistering = whoIsRegistering;
     }
 }
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
@@ -23,6 +23,7 @@ public class OverseasEntitySubmissionDto {
     public static final String MANAGING_OFFICERS_INDIVIDUAL_FIELD = "managing_officers_individual";
     public static final String MANAGING_OFFICERS_CORPORATE_FIELD = "managing_officers_corporate";
     public static final String TRUST_DATA = "trusts";
+    public static final String WHO_IS_REGISTERING = "who_is_registering";
 
     @JsonProperty(PRESENTER_FIELD)
     private PresenterDto presenter;
@@ -57,6 +58,9 @@ public class OverseasEntitySubmissionDto {
     @JsonInclude(NON_NULL)
     @JsonProperty(TRUST_DATA)
     private List<TrustDataDto> trusts;
+
+    @JsonProperty(WHO_IS_REGISTERING)
+    private String whoIsRegistering;
 
     @JsonProperty("links")
     private Map<String, String> links;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceDataBlockValidator.java
```diff
@@ -18,14 +20,36 @@ public class DueDiligenceDataBlockValidator {
...
+         }
+    }
+
+    private boolean incorrectDueDiligenceBlockIsPresent(String whoIsRegistering,
+                                                        DueDiligenceDto dueDiligenceDto,
+                                                        OverseasEntityDueDiligenceDto overseasEntityDueDiligenceDto,
+                                                        Errors errors,
+                                                        String loggingContext) {
+
+        boolean isAgentRegistering = whoIsRegistering.equals(AGENT_REGISTERING);
```
mouhajer-ch, line 44: We need to check if is `agent`, `someone_else` or `null`. This way you can send anything different then agent and it will pass.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -56,12 +57,17 @@ public OverseasEntitySubmissionDtoValidator(EntityDtoValidator entityDtoValidato
     }
 
     public Errors validate(OverseasEntitySubmissionDto overseasEntitySubmissionDto, Errors errors, String loggingContext) {
-        entityDtoValidator.validate(overseasEntitySubmissionDto.getEntity(), errors, loggingContext);
-        presenterDtoValidator.validate(overseasEntitySubmissionDto.getPresenter(), errors, loggingContext);
+        if (UtilsValidators.isNotNull(overseasEntitySubmissionDto.getEntity(), OverseasEntitySubmissionDto.ENTITY_FIELD, errors, loggingContext)) {
```
mouhajer-ch, line 60: Probably we should simplify the following method `validate`, as sonar suggested. If we export related checks, for example for MOs and BOs to a separate private method will reduce the complexity.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
```diff
@@ -23,6 +23,7 @@ public class OverseasEntitySubmissionDto {
     public static final String MANAGING_OFFICERS_INDIVIDUAL_FIELD = "managing_officers_individual";
     public static final String MANAGING_OFFICERS_CORPORATE_FIELD = "managing_officers_corporate";
     public static final String TRUST_DATA = "trusts";
+    public static final String WHO_IS_REGISTERING = "who_is_registering";
```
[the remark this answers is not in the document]
mouhajer-ch: ok, thanks @mattch1 . It will be merged after the SDK is done.

mouhajer-ch (commented):
We need to update the spec as well to match the new object

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceDataBlockValidator.java
mattch1: Have replicated what we do in the web but a null check needs to be added and captured.

mouhajer-ch: Yep, agreed, it is just the check on line 48, we need to confirm that if is not the `agent` has to be `someone_else`.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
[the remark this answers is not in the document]
mouhajer-ch: I suggest to check that only one of them is present, without complicating the logic and saving not important information. So I do not think we need this field

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/DueDiligenceDataBlockValidator.java
mouhajer-ch: Same here, I think we do not need it.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -56,12 +57,17 @@ public OverseasEntitySubmissionDtoValidator(EntityDtoValidator entityDtoValidato
...
+        }
+
+        if (UtilsValidators.isNotNull(overseasEntitySubmissionDto.getPresenter(), OverseasEntitySubmissionDto.PRESENTER_FIELD, errors, loggingContext)) {
+            presenterDtoValidator.validate(overseasEntitySubmissionDto.getPresenter(), errors, loggingContext);
+        }
 
         var dueDiligenceDto = overseasEntitySubmissionDto.getDueDiligence();
         var overseasEntityDueDiligenceDto = overseasEntitySubmissionDto.getOverseasEntityDueDiligence();
-        if (dueDiligenceDataBlockValidator.onlyOneBlockPresent(dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
+        if (dueDiligenceDataBlockValidator.onlyOneCorrectBlockPresent(overseasEntitySubmissionDto.getWhoIsRegistering(), dueDiligenceDto, overseasEntityDueDiligenceDto, errors, loggingContext)) {
```
markpit, line 70: i don't think we need the who_is_registering as the previous check would have been sufficient. It would check that only one of either due diligence or OE due diligence were present. So if only one was present we can assume that is the correct one.

mouhajer-ch (approved):

markpit (approved):

PR #129, opened 2022-10-20
Add cost resource to filing

jphillips2121:
Resolves: UAR-49


emead:
The swagger spec should be updated with the new field

emead (approved):

PR #130, opened 2022-10-24
ROE-1510 Return correct error when aml number is null

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1510

Extra check added to the anti-money laundering number to test whether the field is null so the correct http number and error message are returned.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
@@ -85,7 +85,8 @@ public class DueDiligenceValidator {
         String qualifiedFieldName = getQualifiedFieldName(
                 OverseasEntitySubmissionDto.DUE_DILIGENCE_FIELD,
                 DueDiligenceDto.AML_NUMBER_FIELD);
-        return StringValidators.isLessThanOrEqualToMaxLength(amlNumber, 256, qualifiedFieldName, errors, loggingContext);
+        return StringValidators.isNotBlank(amlNumber, qualifiedFieldName, errors, loggingContext)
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
```diff
@@ -85,7 +85,8 @@ private boolean validateAmlNumber(String amlNumber, Errors errors, String loggin
         String qualifiedFieldName = getQualifiedFieldName(
                 OverseasEntitySubmissionDto.DUE_DILIGENCE_FIELD,
                 DueDiligenceDto.AML_NUMBER_FIELD);
-        return StringValidators.isLessThanOrEqualToMaxLength(amlNumber, 256, qualifiedFieldName, errors, loggingContext);
+        return StringValidators.isNotBlank(amlNumber, qualifiedFieldName, errors, loggingContext)
```
mouhajer-ch, line 88: I think `AML_NUMBER_FIELD` is optional. We should just check the length of the field when is not null, this way you make it required. If it has become required we need to change web and spec

mattch1: We have to fix this bug, the details are here:

https://companieshouse.atlassian.net/browse/ROE-1510

mattch1: We can do that, but the expected result on the ticket will need to be changed.

mattch1: Updated the fix and the the ticket.

mouhajer-ch (approved):

PR #134, opened 2022-10-25
ROE-1447 sanitise exceptions in global exception handler

markpit:
https://companieshouse.atlassian.net/browse/ROE-1447

The GlobalExceptionHandler will now encode and truncate exception messages and stack traces so that if any json parse errors occur, the original value will be encoded and the stack trace truncated before logging.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/exception/GlobalExceptionHandler.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/exception/GlobalExceptionHandler.java
@@ -1,22 +1,48 @@
 package uk.gov.companieshouse.overseasentitiesapi.exception;
 
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.owasp.encoder.Encode;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.WebRequest;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/exception/GlobalExceptionHandler.java
```diff
@@ -1,22 +1,44 @@
...
 
+    @Value("${GLOBAL_EXCEPTION_HANDLER_TRUNCATE_LENGTH_CHARS:15000}")
+    private int truncationLength;
+
     @ExceptionHandler(Exception.class)
-    public ResponseEntity<Object> handleException(Exception ex) {
+    public ResponseEntity<Object> handleException(Exception ex, WebRequest webRequest) {
+        var context = webRequest.getHeader(ERIC_REQUEST_ID_KEY);
+        var sanitisedExceptionMessage = truncate(Encode.forJava(ex.getMessage()));
+        var sanitisedStackTrace = truncate(Encode.forJava(ExceptionUtils.getStackTrace(ex)));
```
mwestacott, line 28: Just wonder if we should have a longer truncation length for the stack trace that may have nested exception stack traces in it? Although if it was ok with your testing in Docker and enough info showed up then that's fine.

markpit: the stack traces in testing were about 11k chars, i'll double check

markpit: the invalid date in managing officer individual dob generates a stack trace of 11.8k. The Identity date in due diligence generates about 11k so seems to be in the region of 11-12k.

mwestacott: 👍 all good then.

mwestacott (approved):

PR #136, opened 2022-11-02
ROE-1543/1603 Include a second nationality in the model and validation

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1543
https://companieshouse.atlassian.net/browse/ROE-1603

Updated:

- model
- validation
- spec
- unit tests

Added a separate field for the second nationality as the emphasis is on "dual" nationality not multiple nationalities it is a single field rather than a list. Followed what we do with former names.

Moreover, as this is not determined by a radio but just an extra field, a simple null check has been added a a condition for checking validation making this an optional field.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/ManagingOfficerIndividualValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/ManagingOfficerIndividualValidator.java
@@ -9,6 +9,7 @@ import uk.gov.companieshouse.overseasentitiesapi.validation.utils.CountryLists;
 import uk.gov.companieshouse.overseasentitiesapi.validation.utils.DateValidators;
 import uk.gov.companieshouse.overseasentitiesapi.validation.utils.StringValidators;
 import uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators;
+import uk.gov.companieshouse.overseasentitiesapi.validation.utils.ValidationMessages;
 import uk.gov.companieshouse.service.rest.err.Errors;
 
 import java.time.LocalDate;
@@ -40,7 +41,7 @@ public class ManagingOfficerIndividualValidator {
             validateDateOfBirth(managingOfficerIndividualDto.getDateOfBirth(), errors, loggingContext);
             validateNationality(managingOfficerIndividualDto.getNationality(), errors, loggingContext);
             if (Objects.nonNull(managingOfficerIndividualDto.getSecondNationality())) {
-                validateSecondNationality(managingOfficerIndividualDto.getSecondNationality(), errors, loggingContext);
+                validateSecondNationality(managingOfficerIndividualDto.getNationality(), managingOfficerIndividualDto.getSecondNationality(), errors, loggingContext);
             }
             validateAddress(ManagingOfficerIndividualDto.USUAL_RESIDENTIAL_ADDRESS_FIELD, managingOfficerIndividualDto.getUsualResidentialAddress(), errors, loggingContext);
             boolean isSameAddressFlagValid = validateIsServiceAddressSameAsUsualResidentialAddress(managingOfficerIndividualDto.getServiceAddressSameAsUsualResidentialAddress(), errors, loggingContext);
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/ManagingOfficerIndividualValidator.java
```diff
@@ -96,6 +100,12 @@ private boolean validateNationality(String nationality, Errors errors, String lo
                 && StringValidators.isValidCharacters(nationality, qualifiedFieldName, errors, loggingContext);
     }
 
+    private boolean validateSecondNationality(String secondNationality, Errors errors, String loggingContext) {
```
markpit, line 103: would you be able to add tests to check the valid chars part of the second nationality validation? in the MO individual and BO individual

mouhajer-ch: Probably it is not part of this story, up to you @mattch1 (not a blocker), but we need to make sure that the user does not select the same nationality as the first one

mattch1: There are tests for that. Yes we should should check that the nationalities are different. Good point.

markpit: it was sonar reporting that a condition was missed, thought it might have been the invalid chars. It could be that there is no test for a valid second nationality, try including it in the 'no errors reported' test

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
```diff
@@ -84,4 +84,13 @@ public static void checkIsEmpty(String toTest, String qualifiedFieldName, Errors
             ApiLogger.infoContext(loggingContext, qualifiedFieldName + " Field should not be populated");
         }
     }
+
+    public static boolean checkIsNotEqual(String nationality, String secondNationality, String errorMsg, String qualifiedFieldName, Errors errors, String loggingContext) {
```
markpit, line 88: the checkIsNotEqual method is generic so should use generic variable and param names.
I'd change String nationality, String secondNationality to String string1, String string2.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/ManagingOfficerIndividualValidator.java
mattch1: More testing with the check equal now up to 100%

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
markpit: if you use StringUtils.equals() in here you won't have to add a null check, otherwise you should check strings are null as could get null pointer exception calling .equals on the string

mattch1: Names updated.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
```diff
@@ -84,4 +84,13 @@ public static void checkIsEmpty(String toTest, String qualifiedFieldName, Errors
             ApiLogger.infoContext(loggingContext, qualifiedFieldName + " Field should not be populated");
         }
     }
+
+    public static boolean checkIsNotEqual(String string1, String string2, String errorMsg, String qualifiedFieldName, Errors errors, String loggingContext) {
```
markpit, line 88: Needs tests added to StringValidatorsTest

markpit: also needs to null check the string if not using StringUtils.equals. null pointer will occur if something passes string1 as null

mouhajer-ch (approved):

PR #140, opened 2022-11-16
ROE-1643 Add 'resume' link to transaction when new submission created

mwestacott:
* Resume link added when initial POST request comes in to controller
* Service changes to create te link if 'save and resume' is active
* Unit test changes

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/Constants.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/Constants.java
@@ -23,6 +23,7 @@ public class Constants {
     public static final String COSTS_URI_SUFFIX = "/costs";
     public static final String TRANSACTIONS_PRIVATE_API_PREFIX = "/private/transactions/";
     public static final String TRANSACTIONS_PUBLIC_API_PREFIX = "/transactions/";
+    public static final String RESUME_JOURNEY_URI_PATTERN = "/register-an-overseas-entity/transactions/%s/overseas-entity/%s";
 
     // Filings
     public static final String FILING_KIND_OVERSEAS_ENTITY = "overseas-entity";
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/Constants.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/Constants.java
@@ -23,7 +23,7 @@ public class Constants {
     public static final String COSTS_URI_SUFFIX = "/costs";
     public static final String TRANSACTIONS_PRIVATE_API_PREFIX = "/private/transactions/";
     public static final String TRANSACTIONS_PUBLIC_API_PREFIX = "/transactions/";
-    public static final String RESUME_JOURNEY_URI_PATTERN = "/register-an-overseas-entity/transactions/%s/overseas-entity/%s";
+    public static final String RESUME_JOURNEY_URI_PATTERN = "/register-an-overseas-entity/transaction/%s/overseas-entity/%s/resume";
 
     // Filings
     public static final String FILING_KIND_OVERSEAS_ENTITY = "overseas-entity";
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/Constants.java
```diff
@@ -23,6 +23,7 @@ private Constants() {
     public static final String COSTS_URI_SUFFIX = "/costs";
     public static final String TRANSACTIONS_PRIVATE_API_PREFIX = "/private/transactions/";
     public static final String TRANSACTIONS_PUBLIC_API_PREFIX = "/transactions/";
+    public static final String RESUME_JOURNEY_URI_PATTERN = "/register-an-overseas-entity/transactions/%s/overseas-entity/%s";
```
mwestacott, line 26: Agreed this format with @mouhajer-ch who is doing the corresponding web task. Does it look ok to everyone?

I originally had '/resume' on the end of it but took this off as we thought it didn't really belong if the web is just doing a GET of a resource with this URL.

markpit: I think we should have the /resume on the end, it will make it easier to identify the route in the web app router and give us a name for a controller eg, resumeController in the web app

markpit: the api get endpoint would be of the format `"/transactions/{transaction_id}/overseas-entity/{overseas_entity_id}`. with no /resume

mouhajer-ch: Hi @markpit, the endpoint is unique on the web and it is easy for the web to identify it, the web endpoint format will be something like `/register-an-overseas-entity/:transaction_id/overseas-entity/:overseas_entity_id`, but if you think the `resume` will give some sort of value we can add it

mouhajer-ch: For the payment callback endpoint we are using `/register-an-overseas-entity/transactions/:transaction_id/overseas-entity/:overseas_entity_id/payment ` with `payment` at the end, I do not mind to use the same logic with `resume` at the end

markpit: yes i think by adding /resume it will just make it clear to the reader exactly what the url is for

mouhajer-ch: So lets add resume then 😄

mwestacott: I can @mouhajer-ch. I had this originally but you suggested 'transactions' plural which I was also then happy with as it matched the example here - https://companieshouse.atlassian.net/wiki/spaces/Arch/pages/3509026858/Transactions+Model#Example-4---Resume-Journey-URI.

I really don't mind. I think if there wasn't that example I'd also go with 'transaction'. @markpit thoughts?

mouhajer-ch: I was referring to the API endpoint, for the web for consistency we are using the same format as for the payment [here](https://github.com/companieshouse/overseas-entities-web/blob/3846ecb38c3f6965e269cabe9c3d6ba8e61a5699/src/config/index.ts#L99)

markpit: the web has used transaction (no s) for payment so i think we should have no 's' for the resume url eg
`/register-an-overseas-entity/transaction/:transactionId/overseas-entity/:overseaEntityId/resume`

the CH apis have 's' so `/transactions/{transaction_id}/overseas-entity/{overseas_entity_id}`. for the api

mwestacott: Alles klar! I will amend the URL.

markpit (approved):

PR #142, opened 2022-11-21
ROE-1645 Add a GET endpoint to retrieve submission data

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1645

New controller, calls service and in the service:

Check user associated with the supplied transaction - added note to the ticket -link to code in transactions api.

Check supplied transaction relates to the supplied OE id as done in the pUT endpoint.

Obtains the submission from mongo and check submission is not empty.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
@@ -241,7 +241,7 @@ public class OverseasEntitiesController {
             ApiLogger.errorContext(requestId, e.getMessage(), e, logMap);
             return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         } catch (Exception e) {
-            ApiLogger.errorContext(requestId, "Error Updating Overseas Entity Submission", e, logMap);
+            ApiLogger.errorContext(requestId, "Error getting Overseas Entity Submission", e, logMap);
             return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
@@ -18,6 +18,7 @@ import org.springframework.web.bind.annotation.RestController;
 import uk.gov.companieshouse.api.model.transaction.Transaction;
 import uk.gov.companieshouse.api.model.validationstatus.ValidationStatusError;
 import uk.gov.companieshouse.api.model.validationstatus.ValidationStatusResponse;
+import uk.gov.companieshouse.overseasentitiesapi.exception.SubmissionNotFoundException;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.OverseasEntitySubmissionDto;
 import uk.gov.companieshouse.overseasentitiesapi.service.OverseasEntitiesService;
 import uk.gov.companieshouse.overseasentitiesapi.utils.ApiLogger;
@@ -137,11 +138,12 @@ public class OverseasEntitiesController {
      * Temporary endpoint for creating an initial OE mongo record with partial data
      * and no validation (to be added). This is to prevent issues with existing POST endpoint
      * that will validate a whole submission.
-     * @param transaction The transaction to be linked to OE submission
+     *
+     * @param transaction                 The transaction to be linked to OE submission
      * @param overseasEntitySubmissionDto The data to store
-     * @param requestId Http request ID, used in logs
-     * @param userId the ERIC user id
-     * @param request the HttpServletRequest
+     * @param requestId                   Http request ID, used in logs
+     * @param userId                      the ERIC user id
+     * @param request                     the HttpServletRequest
      * @return ResponseEntity
      */
     @PostMapping("/start")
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
```diff
@@ -222,6 +224,28 @@ public ResponseEntity<Object> getValidationStatus(
...
+        logMap.put(TRANSACTION_ID_KEY, transaction.getId());
+
+        try {
+            ApiLogger.infoContext(requestId, "Calling service to get the overseas entity submission", logMap);
+            return overseasEntitiesService.getSavedOverseasEntity(transaction, submissionId, requestId);
+        } catch (SubmissionNotFoundException e) {
+            ApiLogger.errorContext(requestId, e.getMessage(), e, logMap);
+            return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+        } catch (Exception e) {
+            ApiLogger.errorContext(requestId, "Error Updating Overseas Entity Submission", e, logMap);
```
markpit, line 244: error message should say 'getting' instead of 'updating'

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
```diff
@@ -133,6 +134,27 @@ public ResponseEntity<Object> updateOverseasEntity(Transaction transaction,
...
 
+    public ResponseEntity<Object> getSavedOverseasEntity(Transaction transaction,
+                                                         String submissionId,
+                                                         String requestId) throws SubmissionNotFoundException {
+        ApiLogger.debugContext(requestId, "Called getOverseasEntity(...)");
+
+        final String submissionUri = getSubmissionUri(transaction.getId(), submissionId);
+
+        if (!transactionUtils.isTransactionLinkedToOverseasEntitySubmission(transaction, submissionUri)) {
+            return ResponseEntity.badRequest().body(String.format(
```
markpit, line 145: i think we should log an error here before we return the badRequest so that if it does happen we can tell what has happened when we look in the logs. So use an errorContext with tranasction_id and submission_id.
we should probably add the logging in to the update method as well (where it also returns badRequest for the same check)

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
```diff
@@ -133,6 +134,27 @@ public ResponseEntity<Object> updateOverseasEntity(Transaction transaction,
         return ResponseEntity.ok().build();
     }
 
+    public ResponseEntity<Object> getSavedOverseasEntity(Transaction transaction,
+                                                         String submissionId,
+                                                         String requestId) throws SubmissionNotFoundException {
+        ApiLogger.debugContext(requestId, "Called getOverseasEntity(...)");
```
mouhajer-ch, line 140: It should be `getSavedOverseasEntity`

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
markpit: actually, the service shouldn't return a response entity, that should be done by the controller, so it would be better here to throw a SubmissionNotLinkedToTransactionException and catch and log that in the controller before returning the response entity from the controller

mattch1: I'm not sure why this wasn't raised when updateOverseasEntity was written. I have based this on whats's there I did think this myself but as it had been previously approved. I just went ahead with it as we had no specific exception for this, which I assumed was by design.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
mattch1: updated this.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesController.java
mattch1: This has been changed.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
markpit: probably just something that was overlooked, better that we spot it now, seems a lot better and more consistent with the new approach, thnaks.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesControllerTest.java
```diff
@@ -613,6 +615,77 @@ void testCreatingANewSaveAndResumeSubmissionIsUnSuccessfulWhenValidationChecksFa
...
+                REQUEST_ID
+        );
+
+        assertEquals(HttpStatus.OK, response.getStatusCode());
+        assertNotNull(response.getBody());
+    }
+
+    @Test
+    void testGetSubmissionIsNotSucessfulWhenBadRequestIsReturned() throws SubmissionNotFoundException, SubmissionNotLinkedToTransactionException {
+        ResponseEntity badRequestResponse = ResponseEntity.badRequest().body(String.format(
```
markpit, line 638: the badRequestResponse isn't used so can be removed

src/test/java/uk/gov/companieshouse/overseasentitiesapi/controller/OverseasEntitiesControllerTest.java
```diff
@@ -613,6 +615,77 @@ void testCreatingANewSaveAndResumeSubmissionIsUnSuccessfulWhenValidationChecksFa
...
+                SUBMISSION_ID,
+                REQUEST_ID
+        );
+
+        assertEquals(HttpStatus.OK, response.getStatusCode());
+        assertNotNull(response.getBody());
+    }
+
+    @Test
+    void testGetSubmissionIsNotSucessfulWhenBadRequestIsReturned() throws SubmissionNotFoundException, SubmissionNotLinkedToTransactionException {
```
markpit, line 637: typo in test name 'Sucessful'

markpit: there's a few with the 'Sucessful' typo

markpit, line 637: i'd change the name of the test as it is testing the response when a SubmissionNotLinked exception is thrown, maybe something like
testGetSubmissionIsNotSuccessfulWhenSubmissionNotLinkedToTransaction

mouhajer-ch (commented):
I forgot to mention that we need to update the spec to reflect the new end-point and related response status

markpit (approved):

PR #143, opened 2022-11-23
ROE-1248 Change email regex to match account.ch.gov.uk

markpit:
https://companieshouse.atlassian.net/browse/ROE-1248

Reverting the email regex back to the OWASP version. This is because the previous regex used is not recognised as valid by javascript in overseas-entities-web


markpit:
This story has been completed by another PR

PR #144, opened 2022-11-24
ROE-1646 Do not allow certain operations on 'Closed' transactions

mwestacott:
https://companieshouse.atlassian.net/browse/ROE-1646

* Required for 'save and resume' but this safety check was missing off OE API end-points that had already been delivered
* Any end-points invoked from the web app (see URL patterns used in the implementation) now check that the associated transaction is not already 'Closed'. If it is, an error is returned to the client
* Achieved using a new 'ProcessingInterceptor'
* More generally, changed log output to use the ERROR level if an interceptor is about to fail the request
* Added unit tests for the new interceptor
* Updated some existing tests to improve on the checks that they were doing

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/FilingInterceptor.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/FilingInterceptor.java
@@ -6,6 +6,10 @@ import uk.gov.companieshouse.overseasentitiesapi.utils.ApiLogger;
 import javax.servlet.http.HttpServletResponse;
 import java.util.HashMap;
 
+/**
+ * A request interceptor class that checks if a request to retrieve filing data is allowed - for this to be true a
+ * transaction must be present in the request attributes and that transaction must be CLOSED.
+ */
 @Component
 public class FilingInterceptor extends AbstractClosedTransactionInterceptor {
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/FilingInterceptor.java
```diff
@@ -1,49 +1,24 @@
...
-
-import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.ERIC_REQUEST_ID_KEY;
-import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.TRANSACTION_ID_KEY;
-import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.TRANSACTION_KEY;
-
-import static uk.gov.companieshouse.api.model.transaction.TransactionStatus.CLOSED;
 
 @Component
-public class FilingInterceptor implements HandlerInterceptor {
+public class FilingInterceptor extends AbstractClosedTransactionInterceptor {
```
markpit, line 10: for the concrete implementation classes FilingInterceptor and ProcessingInterceptor could you add javadoc at class level just to say  eg 'should return TRUE if transaction closed' just to make it clear what the behaviour should be (for anyone reading in future)

mwestacott: Sure. Didn't do this as we generally don't bother with Javadoc but makes sense to me.

markpit (approved):

PR #149, opened 2022-11-29
Roe 1706 public register jurisdiction update

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1706

Added field to dto and dao models and then the validation.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidator.java
@@ -45,9 +45,11 @@ public class EntityDtoValidator {
 
         if (entityDto.isOnRegisterInCountryFormedIn()) {
             validatePublicRegisterName(entityDto.getPublicRegisterName(), errors, loggingContext);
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidator.java
```diff
@@ -132,6 +141,12 @@ private Errors validatePublicRegisterNameIsNotSupplied(String publicRegisterName
         return errors;
     }
 
+    private Errors validatePublicRegisterJurisdictionIsNotSupplied(String publicRegisterJurisdiction, Errors errors, String loggingContext) {
```
markpit, line 144: i would move the position of this jurisdiction validator up in the class , so that it is next to the other jurisdiction validator, then the registration validators will also be next to each other. Just makes it a bit easier to find things in the class.

mattch1: I'm inclined to keep it as it is we have the three validators then the three not supplied validators seems logical either way.

markpit: ok

markpit (approved):

PR #153, opened 2022-12-02
ROE-1675 e-Mail address trimmed when received by the API

mwestacott:
* e-Mail addresses trimmed in the DTO when set
* Mapper tests and mocks updated to ensure that setter methods are trimming the values
* Tested in Postman

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/EntityDto.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/EntityDto.java
@@ -2,6 +2,8 @@ package uk.gov.companieshouse.overseasentitiesapi.model.dto;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+import java.util.Objects;
+
 public class EntityDto {
 
     public static final String NAME_FIELD = "name";
@@ -98,7 +100,7 @@ public class EntityDto {
     }
 
     public void setEmail(String email) {
-        this.email = email.trim();
+        this.email = Objects.isNull(email) ? null : email.trim();
     }
 
     public String getLegalForm() {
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/EntityDto.java
```diff
@@ -98,7 +98,7 @@ public String getEmail() {
     }
 
     public void setEmail(String email) {
-        this.email = email;
+        this.email = email.trim();
```
markpit, line 101: if we receive null, will cause problems - only trim email if not null?

src/test/java/uk/gov/companieshouse/overseasentitiesapi/model/DtoDaoMappingTest.java
```diff
@@ -167,7 +168,7 @@ private OverseasEntitySubmissionDto getOverseasEntitySubmissionDto() {
 
         EntityDto entity = new EntityDto();
         entity.setName("name");
-        entity.setEmail("email");
+        entity.setEmail(" " + EMAIL + " ");
```
markpit, line 171: maybe add comment on line above to highlight the the spaces are used to test the trim functionality (just in case in future someone changes it without realising)

markpit: or create a EMAIL_NEEDS_TRIMMING constant (or better named)

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/EntityDto.java
mwestacott: Oh yes. Just seen another unit test break for this too, which is good. I will amend.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/model/DtoDaoMappingTest.java
mwestacott: Commented added but I've also defined a constant with a more meaningful name (with Javadoc comment).

markpit (approved):

mattch1 (approved):

PR #155, opened 2022-12-09
ROE-1729 updating public register validation and messages

SVarallo:


src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidatorTest.java
```diff
@@ -433,7 +433,7 @@ void testErrorReportedWhenRegistrationNumberFieldContainsInvalidCharacters() {
     @Test
     void testNoErrorReportedWhenPublicRegisterNameAndJurisdictionFieldIsMaxLength() {
         entityDto.setOnRegisterInCountryFormedIn(true);
-        entityDto.setPublicRegisterName(StringUtils.repeat("A", 79));
+        entityDto.setPublicRegisterName(StringUtils.repeat("A", 78));
```
mattch1, line 436: Could we run this with 79 characters as this would be the edge case

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidator.java
```diff
@@ -121,7 +121,7 @@ private boolean validatePublicRegisterName(String publicRegisterName, String pub
         var compoundQualifiedFieldName = String.format("%s and %s", qualifiedFieldNamePublicRegisterName, qualifiedFieldNamePublicRegisterJurisdiction);
 
         return StringValidators.isNotBlank(publicRegisterName, qualifiedFieldNamePublicRegisterName, errors, loggingContext)
-                && StringValidators.isLessThanOrEqualToMaxLength(String.format(CONCATENATED_STRING_FORMAT, publicRegisterName, publicRegisterJurisdiction), 160, compoundQualifiedFieldName, errors, loggingContext)
+                && StringValidators.isLessThanOrEqualToMaxLength(String.format(CONCATENATED_STRING_FORMAT, publicRegisterName, publicRegisterJurisdiction), 159, compoundQualifiedFieldName, errors, loggingContext)
```
markpit, line 124: I think the length check should still be 160, as the name and jurisdiction should be max 159, the comma between them makes it 160.
So i believe we just need to change the error message that gets returned to tell the user they can only enter 159 chars. I would check this with Sarah (who raised the bug)

sarahocarroll: Yes, that's correct - the validation message should read that they are only allowed to enter a max of 159 characters.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidatorTest.java
SVarallo: But wouldn't this move it up to 160 when you include the comma?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidator.java
SVarallo: The error message is generated by the validation itself right? Would that require extra work?

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidatorTest.java
mattch1: yep, thats the edge case as long as it's not >=161 or > 160 it should pass.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/EntityDtoValidator.java
```diff
@@ -131,7 +131,7 @@ private boolean validatePublicRegisterJurisdiction(String publicRegisterName, St
         var compoundQualifiedFieldName = String.format("%s and %s", qualifiedFieldNamePublicRegisterName, qualifiedFieldNamePublicRegisterJurisdiction);
 
         return StringValidators.isNotBlank(publicRegisterJurisdiction, qualifiedFieldNamePublicRegisterJurisdiction, errors, loggingContext)
-                && StringValidators.isLessThanOrEqualToMaxLength(String.format(CONCATENATED_STRING_FORMAT, publicRegisterName, publicRegisterJurisdiction), 160, compoundQualifiedFieldName, errors, loggingContext)
+                && StringValidators.isLessThanOrEqualToMaxLength(String.format(CONCATENATED_STRING_FORMAT, publicRegisterName, publicRegisterJurisdiction), 159, compoundQualifiedFieldName, errors, loggingContext)
```
mattch1, line 134: Just had a look. The concatenated string it supposed to be 160 or less, i.e. with the comma, which means the user input should be 159 to account for it, so I'm not sure this should change as we are testing the string as whole. If we want to test user input only then we'd need to remove the concatenated string format, append the two string and then set it to 159. Any error messages both in the api and the web should specify "user input" when stating 159. @markpit and @sarahocarroll what do you think?

markpit: i'll pull the branch and have a look, i was hoping it would just be a case of updating the error message text.

markpit: Yes, what Matt said should work. instead of using the CONCATENATED_STRING_FORMAT to join the 2 strings together, you could just create a new constant - CONCATENATED_STRING_FORMAT_NO_COMMA = %s%s
That will then check the length of the 2 strings together and use the 159 value like you have done.

mattch1 (approved):

PR #156, opened 2022-12-12
Roe 1632 api make entity name field a string not an object

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1632

Removed separate object block for entity name from the model and removed the separate validation and replaced it with a field in the submission root


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -57,9 +56,9 @@ public Errors validateFull(OverseasEntitySubmissionDto overseasEntitySubmissionD
 
     public Errors validatePartial(OverseasEntitySubmissionDto overseasEntitySubmissionDto, Errors errors, String loggingContext) {
 
-        var entityNameDto = overseasEntitySubmissionDto.getEntityName();
-        if (Objects.nonNull(entityNameDto)) {
-            entityNameDtoValidator.validate(entityNameDto, errors, loggingContext);
+        var entityName = overseasEntitySubmissionDto.getEntityName();
+        if (StringUtils.isNotBlank(entityName)) {
```
mwestacott, line 60: I think this need to be a 'non null' check. isNotBlank() will include an empty string which I think should then flag an error. Need a unit-test for this scenario...

mwestacott (commented):
I think there's a spec change required too.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mattch1: Separated null from empty check and just confined validate method to being an empty check as null has already been tested in both partial and full validation.

markpit (approved):

PR #166, opened 2023-01-23
ROE-1755 adding validation to due diligence to check it matches 'agree'

SVarallo:
https://companieshouse.atlassian.net/browse/ROE-1755

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
@@ -114,6 +114,7 @@ public class DueDiligenceValidator {
                 OverseasEntitySubmissionDto.DUE_DILIGENCE_FIELD,
                 DueDiligenceDto.DILIGENCE_FIELD);
         return StringValidators.isNotBlank(diligence, qualifiedFieldName, errors, loggingContext)
-                && StringValidators.isLessThanOrEqualToMaxLength(diligence, 256, qualifiedFieldName, errors, loggingContext);
+                && StringValidators.isLessThanOrEqualToMaxLength(diligence, 256, qualifiedFieldName, errors, loggingContext)
+                && StringValidators.isValidCharacters(diligence, qualifiedFieldName, errors, loggingContext);
     }
 }
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
@@ -19,6 +19,8 @@ import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.Validat
 @Component
 public class DueDiligenceValidator {
 
+    private static final String DILIGENCE_AGREE = "agree";
+
     private final AddressDtoValidator addressDtoValidator;
 
     @Autowired
@@ -114,7 +116,6 @@ public class DueDiligenceValidator {
         String qualifiedFieldName = getQualifiedFieldName(
                 OverseasEntitySubmissionDto.DUE_DILIGENCE_FIELD,
                 DueDiligenceDto.DILIGENCE_FIELD);
-        String DILIGENCE_AGREE = "agree";
         return StringValidators.isNotBlank(diligence, qualifiedFieldName, errors, loggingContext)
                 && StringValidators.checkIsEqual(diligence, DILIGENCE_AGREE, ValidationMessages.SHOULD_BE_AGREE_ERROR_MESSAGE, qualifiedFieldName, errors, loggingContext);
     }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
```diff
@@ -114,6 +114,7 @@ private boolean validateDiligence(String diligence, Errors errors, String loggin
                 OverseasEntitySubmissionDto.DUE_DILIGENCE_FIELD,
                 DueDiligenceDto.DILIGENCE_FIELD);
         return StringValidators.isNotBlank(diligence, qualifiedFieldName, errors, loggingContext)
-                && StringValidators.isLessThanOrEqualToMaxLength(diligence, 256, qualifiedFieldName, errors, loggingContext);
+                && StringValidators.isLessThanOrEqualToMaxLength(diligence, 256, qualifiedFieldName, errors, loggingContext)
+                && StringValidators.isValidCharacters(diligence, qualifiedFieldName, errors, loggingContext);
```
mouhajer-ch, line 118: If I am correct the only value available for the following field is just the string `agree` as for spec [here](https://github.com/companieshouse/overseas-entities-api/blob/a78508edba73203e54357c96bccd4f3d2bc0a018/spec/schema.json#L589).

SVarallo: Oh yeah I forgot the spec my bad, haven't worked on the api in a while. What does that agree mean?

SVarallo: Oh I see, could there be extra validation required to check it matches the only valid string?

mouhajer-ch: It is just a check box [here](https://github.com/companieshouse/overseas-entities-web/blob/cfa31b37263edaa2b37bb163d5e54b2b9403cbe2/views/due-diligence.html#L75), and if the value is present should be just the `agree` string, even the length should not be necessary.

mwestacott: I agree (sorry, couldn't resist!). So yes, sounds like we should just be checking the value is equal to 'agree', obviously handling null values too.

SVarallo: So will this require a change in the bug or just create a new one entirely?

mouhajer-ch: For sure the fix should be as suggested, probably in the story need some changes on the expected behaviour including the fact that only the `agree` world should be accepted as for implementation and spec.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
```diff
@@ -90,4 +90,13 @@ public static boolean checkIsNotEqual(String string1, String string2, String err
         }
         return true;
     }
+
+    public static boolean checkIsEqual(String string1, String string2, String errorMsg, String qualifiedFieldName, Errors errors, String loggingContext) {
```
mouhajer-ch, line 94: I would have used the `StringValidators.checkIsNotEqual` and obviously negated the result on the `DueDiligenceValidator`  method for the `diligence`. Otherwise re-use it, something like this probably will do
```
public static boolean checkIsEqual(String string1, String string2, String errorMsg, String qualifiedFieldName, Errors errors, String loggingContext) {
    return !StringValidators.checkIsNotEqual(string1, string2, errorMsg, qualifiedFieldName, errors, loggingContext);
}
```

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
```diff
@@ -114,6 +115,6 @@ private boolean validateDiligence(String diligence, Errors errors, String loggin
                 OverseasEntitySubmissionDto.DUE_DILIGENCE_FIELD,
                 DueDiligenceDto.DILIGENCE_FIELD);
         return StringValidators.isNotBlank(diligence, qualifiedFieldName, errors, loggingContext)
-                && StringValidators.isLessThanOrEqualToMaxLength(diligence, 256, qualifiedFieldName, errors, loggingContext);
+                && StringValidators.checkIsEqual(diligence, "agree", ValidationMessages.SHOULD_BE_AGREE_ERROR_MESSAGE, qualifiedFieldName, errors, loggingContext);
```
mouhajer-ch, line 118: I would have create a const variable for `agree`

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/utils/StringValidators.java
mouhajer-ch: Nope. Leave as you have done 👍

mouhajer-ch (approved):

PR #167, opened 2023-01-23
ROE-1925 -  TL - Allow GET calls when transaction 'closed pending payment'

mwestacott:
https://companieshouse.atlassian.net/browse/ROE-1925

* Required for 'Save & Resume' when 'Pay now' link clicked from 'Your filings' screen - web client needs to be able to retrieve OE submission data
* GET calls to retrieve data and also check validation status are now allowed if status of transaction is 'closed pending payment'
* Unit tests added/changed to check new behaviour


mouhajer-ch (approved):
It looks good to me, already tested locally e2e

markpit (approved):

PR #168, opened 2023-01-30
Roe 1946 - TL - handle new entity name block

mattch1:
https://companieshouse.atlassian.net/browse/ROE-1946

- Updated model to change entity name from a field to a block object.
- Added separate verification for entity name.
- Updated spec
- Ensure partial validation returns an error when the entity name field in the block is explicitly null but not when the block is absent.
- Null checks for entity name in the OverseasEntitiesService and the FilingsService.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
@@ -119,7 +119,9 @@ public class FilingsService {
                         new SubmissionNotFoundException(
                                 String.format("Empty submission returned when generating filing for %s", overseasEntityId)));
 
-        data.put(ENTITY_NAME_FIELD, submissionDto.getEntityName().getName());
+        if (Objects.nonNull(submissionDto.getEntityName())) {
+            data.put(ENTITY_NAME_FIELD, submissionDto.getEntityName().getName());
+        }
         data.put(PRESENTER_FIELD, submissionDto.getPresenter());
         data.put(ENTITY_FIELD, submissionDto.getEntity());
         data.put(DUE_DILIGENCE_FIELD, submissionDto.getDueDiligence());
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
@@ -119,7 +119,7 @@ public class FilingsService {
                         new SubmissionNotFoundException(
                                 String.format("Empty submission returned when generating filing for %s", overseasEntityId)));
 
-        data.put(ENTITY_NAME_FIELD, submissionDto.getEntityName());
+        data.put(ENTITY_NAME_FIELD, submissionDto.getEntityName().getName());
         data.put(PRESENTER_FIELD, submissionDto.getPresenter());
         data.put(ENTITY_FIELD, submissionDto.getEntity());
         data.put(DUE_DILIGENCE_FIELD, submissionDto.getDueDiligence());
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
@@ -119,7 +119,9 @@ public class FilingsService {
                         new SubmissionNotFoundException(
                                 String.format("Empty submission returned when generating filing for %s", overseasEntityId)));
 
-        if (Objects.nonNull(submissionDto.getEntityName())) {
+        if (Objects.isNull(submissionDto.getEntityName())) {
+            data.put(ENTITY_NAME_FIELD, null);
+        } else {
             data.put(ENTITY_NAME_FIELD, submissionDto.getEntityName().getName());
         }
         data.put(PRESENTER_FIELD, submissionDto.getPresenter());
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
@@ -119,7 +119,7 @@ public class OverseasEntitiesService {
 
         // Update company name set on the transaction and add a link to our newly created Overseas Entity
         // submission (aka resource) to the transaction (and potentially also a link for the 'resume' journey)
-        updateTransactionWithLinksAndCompanyName(transaction, overseasEntitySubmissionDto.getEntityName(), submissionId,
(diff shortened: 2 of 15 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -119,7 +119,9 @@ private void setSubmissionData(Map<String, Object> data, String overseasEntityId
                         new SubmissionNotFoundException(
                                 String.format("Empty submission returned when generating filing for %s", overseasEntityId)));
 
-        data.put(ENTITY_NAME_FIELD, submissionDto.getEntityName());
+        if (Objects.nonNull(submissionDto.getEntityName())) {
```
markpit, line 122: i'm not sure what will happen if we don't add the ENTITY_NAME_FIELD to the map, so I would keep the map structure the same as previous and populate ENTITY_NAME_FIELD with null if it is missing.
eg
if (Objects.isNull(submissionDto.getEntityName()) {
    data.put(ENTITY_NAME_FIELD, null);
} else {
  data.put(ENTITY_NAME_FIELD, submissionDto.getEntityName().getName());
}

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
```diff
@@ -119,7 +120,7 @@ private ResponseEntity<Object> createOverseasEntity(Transaction transaction,
 
         // Update company name set on the transaction and add a link to our newly created Overseas Entity
         // submission (aka resource) to the transaction (and potentially also a link for the 'resume' journey)
-        updateTransactionWithLinksAndCompanyName(transaction, overseasEntitySubmissionDto.getEntityName(), submissionId,
+        updateTransactionWithLinksAndCompanyName(transaction, overseasEntitySubmissionDto.getEntityName().getName(), submissionId,
```
markpit, line 123: this could throw a null pointer if the submission dto doesn't have EntityName block (it is optional in our partial validation).
I'd check the name before calling updateTransactionWithLinksAndCompanyName
eg
var entityName = null;
if (Objects.nonNull(overseasEntitySubmissionDto.getEntityName())) {
  entityName = overseasEntitySubmissionDto.getEntityName().getName();
}
updateTransactionWithLinksAndCompanyName(transaction, entityName, submissionId,...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
mwestacott: My feeling on this is that we should not worry about the null check and just blow up with a null pointer at this point, if no name section is present (it absolutely should be by this point). Need a chat probably to discuss why...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
mwestacott, line 123: With Postman, can you create an Overseas Entity with no name block? Not saying this code is wrong but I'm interested in the answer and it might inform decisions taken above and below in this PR.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
```diff
@@ -149,7 +150,10 @@ public ResponseEntity<Object> updateOverseasEntity(Transaction transaction,
         updateOverseasEntitySubmissionWithMetaData(overseasEntitySubmissionDao, submissionUri, requestId, userId);
 
         // Update company name set on the transaction, to ensure it matches the value received with this OE submission
-        transaction.setCompanyName(overseasEntitySubmissionDto.getEntityName());
+        if(Objects.nonNull(overseasEntitySubmissionDto.getEntityName())) {
```
mwestacott, line 153: If we perform an update and don't supply a name (block or value), should we be updating the transaction with an empty string for name?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
markpit: actually, my suggestion my also have an edge case.
Instead of my previous suggestion, I'd change the updateTransactionWithLinksAndCompanyName method to take the EntityNameDto as a param instead of the companyNameString
Then, in updateTransactionWithLinksAndCompanyName

change
        transaction.setCompanyName(companyName);

to be 
if (Objects.nonNull(entityNameDto)) {
          transaction.setCompanyName(entityNameDto.getName());
}


this way we will only set the name if it is present.
my previous suggestion would set to null if that block was missing, but the name could have been set by a previous call to PUT, and if a second call to PUT is made and entityNAme block is missing, that is not necessarily an instruction to set the name to null.

I'll include @mwestacott , might need to discuss this

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
mattch1: Update the Filing service to use the if else conditional and pass null when the block is missing.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
mattch1: Updated to set to null first and then update the value if we have the block.

mattch1: Updated to set to null first and then update the value if we have the block.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
mattch1: Updated as above.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingServiceTest.java
```diff
@@ -195,6 +196,32 @@ void testFilingGenerationWhenSuccessfulWithoutTrustsAndWithIdentityChecks(Submis
         checkManagingOfficers(filing);
     }
 
+    @Test
+    void testFilingGenerationAllowedWhenEntityNameBlockToBeNull() throws IOException, URIValidationException, ServiceException, SubmissionNotFoundException, JSONException {
```
mwestacott, line 200: Would rename this method slightly - `testFilingGenerationAllowedWhenEntityNameBlockIsNull`.

mattch1: Renamed this.

mwestacott (approved):

markpit (approved):

PR #170, opened 2023-02-09
uar-298 add feature flag as refunds fix

AndreKramer7654:
UAR-298 https://companieshouse.atlassian.net/browse/UAR-298

Add feature flag for update and return registration costs directly if false.
Temporary fix for refunds calling costs and loading document when schema changed incompatibly.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/CostsController.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/CostsController.java
@@ -1,6 +1,7 @@
 package uk.gov.companieshouse.overseasentitiesapi.controller;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/controller/CostsController.java
```diff
@@ -47,12 +51,21 @@ public ResponseEntity<List<Cost>> getCosts(
...
+                cost = costsService.getCostsForRegistration();
+            }
 
             return ResponseEntity.ok(Collections.singletonList(cost));
         } catch (SubmissionNotFoundException e) {
             ApiLogger.errorContext(requestId, "Error determining submission costs", e, logMap);
             return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
         }
     }
+
```
mwestacott, line 67: We don't need this 'set' method, you can just make use of ReflectionTestUtils in the unit-test to set the value to what you want. E.g.

        ReflectionTestUtils.setField(costController, "isRoeUpdateEnabled", value);

See the `OverseasEntitiesControllerTest` for a working example.

AndreKramer7654: Happy to change that but not too fond of using reflection for this but as it's already used.

mwestacott (approved):

markpit (approved):

PR #171, opened 2023-02-10
Uar 298 add feature flag with schema mod fix

AndreKramer7654:
https://companieshouse.atlassian.net/browse/UAR-298

Use a separate repository to read only the entity_number (if present). As this is optional, avoids any schema changes in rest of doc impacting the cost endpoint which should then work for any doc (as long it is in mongo / retrievable).

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
@@ -68,19 +68,17 @@ public class OverseasEntitiesService {
     }
 
     SubmissionType getSubmissionType(String requestId, String overseasEntityId) throws SubmissionNotFoundException {
-        var submissionEntityNumberOpt = getEntityNumberFromOverseasEntitySubmission(overseasEntityId);
+        var submissionEntityNumberOpt = getEntityNumberFromOverseasEntitySubmission(
+                overseasEntityId);
         if (submissionEntityNumberOpt.isEmpty()) {
-            throw new SubmissionNotFoundException("Can not determine submission type");
-        }
-
-        String entityNumber = submissionEntityNumberOpt.get();
-        if (StringUtils.isNotBlank(entityNumber)) {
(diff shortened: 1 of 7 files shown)


olaych (approved):

bantao-ch (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
```diff
@@ -230,6 +235,16 @@ private void updateTransactionWithLinksAndCompanyName(Transaction transaction,
         transactionService.updateTransaction(transaction, loggingContext);
     }
 
+    public Optional<String> getEntityNumberFromOverseasEntitySubmission(String submissionId) {
```
mwestacott, line 238: I would have implemented this method differently - the optionality of an entity number being returned would be based upon whether this was an UPDATE submission (so in this case present) or a REGISTRATION submission (not present). So line 242 would be changed to check for null (using `Objects.isNull()`) and return either an `Optional.of(entityNumber)` or `Optional.empty()`.

Line 244 would then change to throw `new SubmissionNotFoundException("Can not determine submission type")`, precluding the need to do this in the `getSubmissionType()` method above. The logic in that method would then also change slightly.

mwestacott (commented):
Not sure we need to do this or merge this right now, at least not before discussing the planned approach/design for dealing with model changes.

I think we might discuss this further at the Architectural Design Catchup meeting tomorrow.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/OverseasEntitiesService.java
AndreKramer7654: Yeah, I did not want to throw from this "get" method but have now followed your suggestion.

AndreKramer7654:
Yes, up for discussion of how we allow for schema changes / need to protect in future.

AndreKramer7654:
No longer likely to be needed

PR #201, opened 2023-03-30
ROE-2092 -TL- Partial Validation null check required for statement enum

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2092

### Change description

The 500 error was a caused by Jackson trying to change an empty string into an enum value.

Added Json creator annotation to the BeneficialOwnersStatementType enum's find method to return null if the enum is not matched when jackson tries to convert the empty string to an enum value, as advised here:
https://stackoverflow.com/questions/51406816/jackson-cannot-convert-empty-string-value-to-enum

New wrapper method for partial validation that checks whether the statement enum is null and returns the error expected on the ticket.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/BeneficialOwnersStatementType.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/BeneficialOwnersStatementType.java
@@ -1,5 +1,7 @@
 package uk.gov.companieshouse.overseasentitiesapi.model;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+
 public enum BeneficialOwnersStatementType {
 
     ALL_IDENTIFIED_ALL_DETAILS("all_identified_all_details"),
@@ -11,7 +13,7 @@ public enum BeneficialOwnersStatementType {
    BeneficialOwnersStatementType(String beneficialOwnersStatement) {
         this.beneficialOwnersStatement = beneficialOwnersStatement;
     }
-
+    @JsonCreator
     public static BeneficialOwnersStatementType findByBeneficialOwnersStatementTypeString(String beneficialOwnersStatement) {
         for (BeneficialOwnersStatementType type: values()) {
             if(type.beneficialOwnersStatement.equals(beneficialOwnersStatement)) {
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
@@ -140,7 +140,7 @@ public class OverseasEntitySubmissionDtoValidator {
 
         errors = validatePartialCommonDetails(overseasEntitySubmissionDto, errors, loggingContext);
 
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/BeneficialOwnersStatementType.java
```diff
@@ -11,7 +13,7 @@ public enum BeneficialOwnersStatementType {
    BeneficialOwnersStatementType(String beneficialOwnersStatement) {
         this.beneficialOwnersStatement = beneficialOwnersStatement;
     }
-
+    @JsonCreator
```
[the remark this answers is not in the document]
mattch1: No they don't fail.

mattch1:
Tested with PUT and get the 400 error back
Now getting 400 back also with POST methods with and without /start appended.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/BeneficialOwnersStatementType.java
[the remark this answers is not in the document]
mwestacott: Ok. Just need to make sure that this change has been 'exercised' by your Postman testing.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/BeneficialOwnersStatementType.java
```diff
@@ -11,13 +13,13 @@ public enum BeneficialOwnersStatementType {
...
-
-    public static BeneficialOwnersStatementType findByBeneficialOwnersStatementTypeString(String beneficialOwnersStatement) {
-        for (BeneficialOwnersStatementType type: values()) {
-            if(type.beneficialOwnersStatement.equals(beneficialOwnersStatement)) {
-                return type;
-            }
-        }
-        return null;
-    }
+   @JsonCreator
```
mwestacott, line 16: Blank line missing above the annotation and I think that @markpit also suggested a comment here as to what this achieves.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/BeneficialOwnersStatementType.java
```diff
@@ -12,12 +14,18 @@ public enum BeneficialOwnersStatementType {
...
-    public static BeneficialOwnersStatementType findByBeneficialOwnersStatementTypeString(String beneficialOwnersStatement) {
-        for (BeneficialOwnersStatementType type: values()) {
-            if(type.beneficialOwnersStatement.equals(beneficialOwnersStatement)) {
-                return type;
-            }
-        }
-        return null;
-    }
+    /**
+     * Json creator allows non-matching enum values to be interpreted by the JSON serialiser as null
```
mwestacott, line 18: Thanks for adding this. Comment not exactly correct though - the annotation simply says this method will be used when interpreting the values. Whether null is returned, or something else, is down to the method implementation.

Sorry to be pedantic, but I am like a human code compiler of plain English... 😆 .

mattch1: Updated the comment to refer to the method.

markpit (approved):

PR #227, opened 2023-05-03
ROE-2172 - TL - Do not validate identity date for partial validation

mattch1:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2172

### Change description

Split due diligence into verification with and without identity date and linked it to full and partial validation respectively.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
@@ -28,8 +28,13 @@ public class DueDiligenceValidator {
         this.addressDtoValidator = addressDtoValidator;
     }
 
-    public Errors validate(DueDiligenceDto dueDiligenceDto, Errors errors, String loggingContext) {
+    public Errors validateWithIdentityDate(DueDiligenceDto dueDiligenceDto, Errors errors, String loggingContext) {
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/DueDiligenceValidator.java
```diff
@@ -28,8 +28,13 @@ public DueDiligenceValidator(AddressDtoValidator addressDtoValidator) {
         this.addressDtoValidator = addressDtoValidator;
     }
 
-    public Errors validate(DueDiligenceDto dueDiligenceDto, Errors errors, String loggingContext) {
+    public Errors validateWithIdentityDate(DueDiligenceDto dueDiligenceDto, Errors errors, String loggingContext) {
```
markpit, line 31: i think it makes better sense to keep 'validate' as the full 'everything' validator and have a 'validateWithoutIdentityDate' so switching the naming around but having 'without' instead of 'with'.

mattch1: Swapped names around

mwestacott (commented):
Looks good. Assume you've tested this in Docker with 'save and resume', modifying the due diligence date in Mongo to more than 3 months in the past before resuming and checking you can move off the 'entity name' screen?

mwestacott (approved):

PR #235, opened 2023-05-10
Feature/uar 413 collate public and private data

RobertBromfield:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-399

### Change description

Created class PublicPrivateDataCombiner which contains HashMaps of the OE, MO, BO data 

Here the keys are a hashed ID and the value is a Pair containing the Public Data on the right and the Private Data on the Left 

#### Before Committing 

Please check that the pom.xml is up to date this PR relies on changes in other repos still in the pipeline

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/HashHelper.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/HashHelper.java
@@ -1,0 +1,29 @@
+package uk.gov.companieshouse.overseasentitiesapi.utils;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Base64;
+import org.springframework.beans.factory.annotation.Value;
+
+public class HashHelper {
+    @Value("${PUBLIC_API_IDENTITY_HASH_SALT:''}")
+    private String salt;
+
+    public String encode(String plain) throws NoSuchAlgorithmException {
+        return base64Encode(sha1Digest(plain)).replace("=", "");
+    }
+
+    private String base64Encode(final byte[] bytes) {
+        return Base64.getUrlEncoder().encodeToString(bytes);
+    }
+
+    private byte[] sha1Digest(final String plain) throws NoSuchAlgorithmException {
+        return MessageDigest.getInstance("SHA-1").digest((plain + salt).getBytes(
+                StandardCharsets.UTF_8));
+    }
+
+    public void setSalt(String salt) {
+        this.salt = salt;
+    }
+}
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/HashHelper.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/HashHelper.java
@@ -4,8 +4,6 @@ import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
-import org.springframework.beans.factory.annotation.Value;
-import org.springframework.stereotype.Component;
 
 public class HashHelper {
 
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/HashHelper.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/HashHelper.java
@@ -5,10 +5,15 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Base64;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
 
 public class HashHelper {
-    @Value("${PUBLIC_API_IDENTITY_HASH_SALT:''}")
+
     private String salt;
+
+    public HashHelper(String salt) {
+        this.salt = salt;
+    }
 
     public String encode(String plain) throws NoSuchAlgorithmException {
         return base64Encode(sha1Digest(plain)).replace("=", "");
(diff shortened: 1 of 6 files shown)


vgummadala (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Puts a value into a map, either in the left or right of a pair depending on the leftRight
+   *
+   * @param map      Map to put the value into
+   * @param key      Key to put the value under
+   * @param value    Value to put into the Pair in the map
+   * @param position Whether to put the value in the left or right of the pair
+   */
+  private <L, R> void putInPairMap(Map<String, Pair<L, R>> map, String key, Object value,
```
AndreKramer7654, line 111: Can we have two methods to addPrivateData and addPublicData that don't use casts?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+   * Hashes an ID using SHA-1 algorithm before Base 64 Encoding
+   *
+   * @param id The ID to hash
+   * @return Hashed ID using SHA-1 algorithm and Base 64 Encoded
+   * @throws NoSuchAlgorithmException
+   */
+  private String getHashedId(String id) throws NoSuchAlgorithmException {
+    var hashHelper = new HashHelper(this.salt);
```
AndreKramer7654, line 130: Keep hashHelper as private member and just call it to do the work when needed. No need for this private method.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -0,0 +1,136 @@
...
+import uk.gov.companieshouse.overseasentitiesapi.service.PrivateDataRetrievalService;
+import uk.gov.companieshouse.overseasentitiesapi.service.PublicDataRetrievalService;
+
+public class PublicPrivateDataCombiner {
+
+  private final PublicDataRetrievalService publicDataRetrievalService;
+  private final PrivateDataRetrievalService privateDataRetrievalService;
+  private final String salt;
+
+  public PublicPrivateDataCombiner(final PublicDataRetrievalService publicDataRetrievalService,
```
AndreKramer7654, line 22: Any pair in the map which only has private data is a consistency error between the apis. I wonder should we check for this case and print errors and remove them from the map. A pcs/officer having only public data is also a consistency error (we should warn) but we have no choice but to work with that. The combining code is going to have to be very careful with this incomplete pair though!

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/HashHelper.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.companieshouse.overseasentitiesapi.utils;
```
AndreKramer7654, line 1: This is now merged so can update PR.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
RobertBromfield: I feel like that violates DRY Principles if I'm honest.
The way it's written currently there is no chance of their being an issue with casting

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -0,0 +1,136 @@
...
+    this.salt = salt;
+  }
+
+  /**
+   * Returns an instance of Pair containing the public data on the right and private data on the
+   * left
+   *
+   * @return Pair of the public data on the right and private data on the left
+   */
+  public Pair<CompanyProfileApi, OverseasEntityDataApi> buildMergedOverseasEntityDataPair() {
```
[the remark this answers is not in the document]
RobertBromfield: I'm not sure what you mean.

When I spoke to you previously you were ok with using a Pair and the first half of this comment says that you still are.

I'm not sure how creating a data class will help except make the code more verbose.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/HashHelper.java
RobertBromfield: Unfortunately because that branch was squashed there doesn't seem to be a way to unpick the previous commits from the one in this branch 

I've made a new branch and will be doing these comments in there tomorrow

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
AndreKramer7654: I think we need to take the error checking into a new story. While it might be more efficient to dedect missing public/private BO/MOs there are other cases we also need to consider.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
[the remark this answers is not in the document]
AndreKramer7654: I'd like it to be type safe though - without casts.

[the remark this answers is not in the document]
RobertBromfield: Ok but that's an entirely different conversation 

I can use get rid of the putInPairMap method which uses casts and just repeat the code in `buildMergedManagingOfficerDataMap()` and `buildMergedBeneficialOwnerDataMap()`

Like I said I don't think it's necessary but it's a different conversation to getting rid of Pairs entirely 

To do that would add an extra day or two to the sprint and hold up other tickets which seems very unnecessary

PR #240, opened 2023-05-16
UAR-494: return updateSubmission model from getFiling()

bmcgreevyCH:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-494

### Change description
Return a new updateSubmission model from getFiling()


src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +93,116 @@ public FilingsService(
...
+      publicDataRetrievalService.initialisePublicData(submissionDto.getEntityNumber(), passThroughTokenHeader);
       privateDataRetrievalService.initialisePrivateData(submissionDto.getEntityNumber());
+      UpdateSubmission updateSubmission = new UpdateSubmission();
+      setUpdateSubmissionData(updateSubmission, submissionDto, logMap);
+      Map<String, Object> updateSubmissionMap = objectMapper.convertValue(updateSubmission, new TypeReference<Map<String, Object>>() {});
+      filing.setData(updateSubmissionMap);
+      filing.setKind(FILING_KIND_OVERSEAS_ENTITY_UPDATE);
+    } else {
+      setSubmissionData(userSubmission, submissionDto, logMap);
+      filing.setData(userSubmission);
```
AndreKramer7654, line 137: where is filing.setKind() called for registration? I'm a bit concerned that we've changed the flow for registration - please make sure nothing is changed on that path.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +93,116 @@ public FilingsService(
...
+                    ));
 
     if (submissionDto.isForUpdate()) {
-      publicDataRetrievalService.initialisePublicData(
-          submissionDto.getEntityNumber(), passThroughTokenHeader);
+      publicDataRetrievalService.initialisePublicData(submissionDto.getEntityNumber(), passThroughTokenHeader);
       privateDataRetrievalService.initialisePrivateData(submissionDto.getEntityNumber());
+      UpdateSubmission updateSubmission = new UpdateSubmission();
+      setUpdateSubmissionData(updateSubmission, submissionDto, logMap);
+      Map<String, Object> updateSubmissionMap = objectMapper.convertValue(updateSubmission, new TypeReference<Map<String, Object>>() {});
```
AndreKramer7654, line 132: userSubmission - now used only for registration is just a new HashMap(). So why is updateSubmissioMap using an objectMapper? Can't we just use updateSubmission for both update and registration setData()?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +93,116 @@ public FilingsService(
...
+          String requestId,
+          String overseasEntityId,
+          Transaction transaction,
+          String passThroughTokenHeader)
+          throws SubmissionNotFoundException, ServiceException {
     var logMap = new HashMap<String, Object>();
     logMap.put(OVERSEAS_ENTITY_ID_KEY, overseasEntityId);
     logMap.put(TRANSACTION_ID_KEY, transaction.getId());
 
     Map<String, Object> userSubmission = new HashMap<>();
```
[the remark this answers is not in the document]
bmcgreevyCH: (This PR is unusually hard to read, the code base is more straightforward)
Its set in setUpdateSubmissionData() below.
updateSubmission.setUserSubmission(submissionDto);

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +93,116 @@ public FilingsService(
...
 
+//    OverseasEntitySubmissionDto submissionDto = setSubmissionData(userSubmission, overseasEntityId, logMap);
     OverseasEntitySubmissionDto submissionDto =
-        setSubmissionData(userSubmission, overseasEntityId, logMap);
-    setPaymentData(userSubmission, transaction, passThroughTokenHeader, logMap);
+            overseasEntitiesService.getOverseasEntitySubmission(overseasEntityId)
```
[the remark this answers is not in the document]
bmcgreevyCH: setSubmissionData() is now for registration flow only.
setUpdateSubmissionData() is for updates.
They both populate different models in the end. Registration converts submissionDto to a Map to put in the **data** Map object of FilingApi, whereas Update just puts submissionDto straight into UpdateSubmission.userSubmission because they are both of type OverseasEntitySubmissionDto, so no conversion needed when its an update.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
bmcgreevyCH: It won't let me put updateSubmission into the data Object because they are mismatching types.
image

AndreKramer7654: But everything is an object!

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +93,117 @@ public FilingsService(
...
+  private void setUpdateSubmissionData(UpdateSubmission updateSubmission, OverseasEntitySubmissionDto submissionDto, Map<String, Object> logMap) {
 
-    OverseasEntitySubmissionDto submissionDto =
-        submissionOpt.orElseThrow(
-            () ->
-                new SubmissionNotFoundException(
-                    String.format(
-                        "Empty submission returned when generating filing for %s",
-                        overseasEntityId)));
+    // Populate relevant fields in UpdateSubmission object //
```
AndreKramer7654, line 156: I think the confusing was caused by setUpdateSubmissionData().  It's now just a one line method that is miss-named if you compare it to setSubmissionData() - which does the inserts into the data map. I'd inline setUpdateSubmissionData() but maybe refactor out a method for the update parts if that makes the setFilingApiData() shorter.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +93,117 @@ public FilingsService(
...
     }
 
-    filing.setData(userSubmission);
     if (overseasEntitiesService.isSubmissionAnUpdate(requestId, overseasEntityId)) {
       filing.setCost(updateCostAmount);
     } else {
       filing.setCost(registerCostAmount);
     }
 
+    setPaymentData(userSubmission, transaction, passThroughTokenHeader, logMap);
```
AndreKramer7654, line 145: I don't think you are adding the payment data in the update case as the map (first arg) is only for registration.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
bmcgreevyCH: > But everything is an Object!

Well that is true. I just got it into my head that I wanted to use matching objects, lol. Okay I'll go with the userSubmission Map for Updates also, to avoid refactoring any current code for Registration.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
bmcgreevyCH: Agreed.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
bmcgreevyCH: Yeah that line of code was more of a placeholder, I've removed it to avoid confusion. This method is currently just a shell, its for the other tickets to fill and update their models from, so setFilingApiData() doesn't get cluttered.

AndreKramer7654: I'd still inline it and let the follow on PR refactor. Probably a bigger piece of setFilingApiData() could be broken out - with a more sensible name in next PR.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +93,112 @@ public FilingsService(
...
+                            String.format("Empty submission returned when generating filing for %s", overseasEntityId)
+                    ));
 
     if (submissionDto.isForUpdate()) {
-      publicDataRetrievalService.initialisePublicData(
-          submissionDto.getEntityNumber(), passThroughTokenHeader);
-      privateDataRetrievalService.initialisePrivateData(submissionDto.getEntityNumber());
+      getPublicAndPrivateData(submissionDto.getEntityNumber(), passThroughTokenHeader);
+      setSubmissionDataForUpdates(userSubmission, submissionDto, logMap);
+      filing.setData(userSubmission);
```
AndreKramer7654, line 129: this line is common to both branches of if (lines 129, 133)

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +93,112 @@ public FilingsService(
...
+  }
 
-    OverseasEntitySubmissionDto submissionDto =
-        submissionOpt.orElseThrow(
-            () ->
-                new SubmissionNotFoundException(
-                    String.format(
-                        "Empty submission returned when generating filing for %s",
-                        overseasEntityId)));
+  private void setSubmissionDataForUpdates(Map<String, Object> data, OverseasEntitySubmissionDto submissionDto, Map<String, Object> logMap) {
```
AndreKramer7654, line 152: This should also get a different arg or extra arg - the actual UpdateSubmission model. I also find the naming of this method confusing as is. Can we just replace where this is called by a code comment and remove this empty method. I've not checked SonarQube but would expect it to complain.

bmcgreevyCH: Agreed. Comment added to replace method call.

AndreKramer7654 (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -92,113 +91,107 @@ public FilingsService(
...
+                    .orElseThrow(() -> new SubmissionNotFoundException(
+                            String.format("Empty submission returned when generating filing for %s", overseasEntityId)
+                    ));
 
     if (submissionDto.isForUpdate()) {
-      publicDataRetrievalService.initialisePublicData(
-          submissionDto.getEntityNumber(), passThroughTokenHeader);
-      privateDataRetrievalService.initialisePrivateData(submissionDto.getEntityNumber());
+      getPublicAndPrivateData(submissionDto.getEntityNumber(), passThroughTokenHeader);
+      //Call a method to populate UpdateSubmission
```
n-r-b, line 126: Is this comment related to another ticket or can it be removed?

AndreKramer7654: For now we could instead do:
// To be replaced with UpdateSubmission model in JIRA number:
setSubmissionData(userSubmission, submissionDto, logMap); 
and re-enable the test.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
vgummadala, line 1: Line 253: Is it possible to put a ternary operator ? 

      if (trustData.size() > 1) {
        throw new ServiceException("There is more than one trust with the ID: " + trustId);
      }
      // If there are is no trust with that ID throw an error
      if (trustData.isEmpty()) {
        throw new ServiceException("There are no trusts for the ID: " + trustId);
      }

      trustsDataForBO.add(trustData.get(0));

vgummadala (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
AndreKramer7654: let's not re-write code outside of our story. Already quite a bit of re-factoring here.

PR #241, opened 2023-05-16
feat: UAR-413 Added Logging

RobertBromfield:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-413

### Work checklist

### Merge instructions

```
```


vgummadala (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -124,6 +124,17 @@ private void setFilingApiData(
       publicDataRetrievalService.initialisePublicData(
           submissionDto.getEntityNumber(), passThroughTokenHeader);
       privateDataRetrievalService.initialisePrivateData(submissionDto.getEntityNumber());
+
+      var publicPrivateDataCombiner = new PublicPrivateDataCombiner(
```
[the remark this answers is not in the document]
AndreKramer7654: Use @Value to get the env value and @Autowire as component.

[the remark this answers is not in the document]
AndreKramer7654: or at least pass in from @Value("PUBLIC_API_IDENTITY_HASH_SALT")

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -108,11 +120,15 @@ public Map<String, Pair<PscApi, PrivateBoDataApi>> buildMergedBeneficialOwnerDat
    */
   public Map<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>> buildMergedManagingOfficerDataMap()
       throws ServiceException {
-    var combineMO = new HashMap<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>>();
+    this.combineMO = new HashMap<>();
 
     var privateMOs = privateDataRetrievalService.getManagingOfficerData();
     var publicOfficers = publicDataRetrievalService.getOfficers();
 
+    if (privateMOs == null || privateMOs.getManagingOfficerData() == null) {
```
AndreKramer7654, line 128: always fetch data

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -151,4 +167,61 @@ public Map<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>> buildMergedM
 
     return combineMO;
   }
+
+  public String logCollatedData() {
+    var result = new StringBuilder("{");
+
+    result.append("\"Managing Officer Data\": [");
+
+    Iterator<Entry<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>>> moIterator = this.combineMO.entrySet().iterator();
```
AndreKramer7654, line 176: for (Entry<> item : this.combineMO) {
}

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -151,4 +167,61 @@ public Map<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>> buildMergedM
...
+        result.append(", ");
+      }
+    }
+    result.append("],");
+
+    result.append("\"Beneficial Owner Data\": [");
+    var boEntries = this.combineBO.entrySet();
+
+    Iterator<Entry<String, Pair<PscApi, PrivateBoDataApi>>> boIterator = boEntries.iterator();
+    while(boIterator.hasNext()) {
```
AndreKramer7654, line 198: foreach

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -151,4 +167,61 @@ public Map<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>> buildMergedM
...
+    result.append("\"Managing Officer Data\": [");
+
+    Iterator<Entry<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>>> moIterator = this.combineMO.entrySet().iterator();
+    while (moIterator.hasNext()) {
+      Entry<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>> officer = moIterator.next();
+      String[] linkAsArray = officer.getValue().getLeft().getLinks().getSelf().split("/");
+      String publicMoId = linkAsArray[linkAsArray.length - 1];
+
+      result.append("{");
+      result.append("\"Map Key\": \"" + officer.getKey() + "\", ");
```
AndreKramer7654, line 183: Surely the Map Key == the public hashed id - no need to log both

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -20,6 +22,10 @@ public class PublicPrivateDataCombiner {
   private final PrivateDataRetrievalService privateDataRetrievalService;
   private final HashHelper hashHelper;
 
+  private Pair<CompanyProfileApi, OverseasEntityDataApi> combinedOE;
+  private Map<String, Pair<PscApi, PrivateBoDataApi>> combineBO;
```
AndreKramer7654, line 26: combinedBOs
combinedMOs

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
AndreKramer7654: I'd remove this if () return
and handle the left and/or right being null in the log method

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
RobertBromfield: So I originally had a foreach but unfortunately you need Iterator in an instance to check if it's the last one

Else it prints a comma at the end

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
RobertBromfield: I'm unsure what this comment means 

This if statement was put in to prevent NullPointerExceptions further on in the code

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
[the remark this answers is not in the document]
RobertBromfield: So I originally had it as an `@value` when I was doing the first implementation 

The DtoModelChangeTest.testCanGenerateDtoModelFromAllJsonVersions tests to fail.

These look to be something CompaniesHouse wrote and therefore something I didn't want to change.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
RobertBromfield: Well it should if it's working correctly I just thought it might be good to log both in case there is an issue somewhere 

I can remove it though

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
RobertBromfield: So the test testFilingGenerationWhenSuccessfulWithoutTrustsAndWithIdentityChecksForUpdate causes this to have a null pointer error if there are no null checks 

If you like I can wrap the for loops in the if statement instead of having it return a null

RobertBromfield: So the test FilingServiceTest.testFilingGenerationWhenSuccessfulWithoutTrustsAndWithIdentityChecksForUpdate causes this to have a NullPointer Exception on the for loop of buildMergedManagingOfficerDataMap

Because of that, there does need to be some kind of if statement to check for null

I can wrap the for loop in an if statement if you think that would be better?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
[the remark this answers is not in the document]
AndreKramer7654: Maybe you did not have the env var in your env? The docker_chs_development repo should have them.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
AndreKramer7654: Please mock getManagingOfficersData() to return a non-null value. We should not be fixing code to pass tests. We would not expect to return nulls when not in test?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
AndreKramer7654: Add the , before you add an item - except for first when none previously added. Instead of adding after if not last. I just think a foreach would be a lot clearer.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
AndreKramer7654: Add the , before you add an item - except for first when none previously added. Instead of adding after if not last. I just think a foreach would be a lot clearer.

AndreKramer7654: It just does not look like modern Java to use an iterator directly. A trick is to add the "," and then remove last after loop if endsin. Or use streams, flatMap and limit,

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -124,6 +127,16 @@ private void setFilingApiData(
...
```
AndreKramer7654, line 131: In future I would like to see Spring components and auto-wiring being used.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -20,6 +20,10 @@ public class PublicPrivateDataCombiner {
   private final PrivateDataRetrievalService privateDataRetrievalService;
   private final HashHelper hashHelper;
 
+  private Pair<CompanyProfileApi, OverseasEntityDataApi> combinedOEs;
+  private Map<String, Pair<PscApi, PrivateBoDataApi>> combineBOs;
```
AndreKramer7654, line 24: combinedBos, combinedMOs

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -151,4 +157,61 @@ public Map<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>> buildMergedM
...
+    var first = true;
+
+    for (var officer : this.combineMO.entrySet()) {
+      if (first) {
+        first = false;
+      } else {
+        result.append(", ");
+      }
+
+      String[] linkAsArray = officer.getValue().getLeft().getLinks().getSelf().split("/");
```
AndreKramer7654, line 174: what if no left was found? Or no right was found?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
RobertBromfield: You're saying in the future so you're ok with it how it is now?

Only Swetha and others are waiting for this and I really don't think another substantial change is wise

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
```diff
@@ -151,4 +157,61 @@ public Map<String, Pair<CompanyOfficerApi, ManagingOfficerDataApi>> buildMergedM
...
+
+    first = true;
+    for (var owner : this.combineBOs.entrySet()) {
+      if (first) {
+        first = false;
+      } else {
+        result.append(", ");
+      }
+
+      String[] linkAsArray = owner.getValue().getLeft().getLinks().getSelf().split("/");
```
AndreKramer7654, line 195: Possible this will break horribly if we don't find left or right

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
AndreKramer7654: It's OK for now.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PublicPrivateDataCombiner.java
RobertBromfield: Yes this is what the null checks earlier were for 

Yesterday you said this scenario would be handled in another ticket wouldn't it be best to leave this till then?

src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingServiceTest.java
```diff
@@ -201,6 +306,7 @@ void testFilingGenerationWhenSuccessfulWithoutTrustsAndWithIdentityChecksForRegi
 //    void testFilingGenerationWhenSuccessfulWithoutTrustsAndWithIdentityChecksForUpdate() throws SubmissionNotFoundException, ServiceException, IOException, URIValidationException {
```
AndreKramer7654, line 306: @Disabled
or remove code

AndreKramer7654 (approved):

PR #264, opened 2023-06-05
fix/UAR-580 Null Pointer Exception fix

RobertBromfield:
### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
@@ -23,6 +23,13 @@ public class ChangeManager<T, L, R> {
    * @param objectToAddChanges The object containing the data that has been changed
    */
   public ChangeManager(T objectToAddChanges, Pair<L, R> pair) {
+    if (objectToAddChanges == null) {
+      throw new IllegalArgumentException("Invalid input parameter: objectToAddChanges is null.");
+    }
+    if (pair == null) {
+      throw new IllegalArgumentException("Invalid input parameter: pair is null.");
+    }
+
     this.objectToAddChanges = objectToAddChanges;
     this.leftPairData = pair.getLeft();
     this.rightPairData = pair.getRight();
@@ -52,23 +59,32 @@ public class ChangeManager<T, L, R> {
       BiPredicate<P, C> equalityPredicate,
       BiConsumer<T, S> dataSetter) {
 
-    var convertedProposedData = proposedConverter.apply(proposedData);
-
-    if (leftOrRightDataFromPair == null && objectToAddChanges == null) {
+    if (isEmpty(proposedData)) {
       return false;
     }
+    if (currentDataGetter == null) {
+      throw new IllegalArgumentException("Invalid input parameter: currentDataGetter is null.");
+    }
+    if (equalityPredicate == null) {
+      throw new IllegalArgumentException("Invalid input parameter: equalityPredicate is null.");
+    }
+    if (proposedConverter == null) {
+      throw new IllegalArgumentException("Invalid input parameter: proposedConverter is null.");
+    }
+    if (dataSetter == null) {
+      throw new IllegalArgumentException("Invalid input parameter: dataSetter is null.");
+    }
 
-    if (
-        !isEmpty(convertedProposedData) &&
-            (leftOrRightDataFromPair == null ||
-                !equalityPredicate.test(proposedData,
-                    currentDataGetter.apply(leftOrRightDataFromPair)))
-    ) {
+    var convertedProposedData = proposedConverter.apply(proposedData);
+    C currentData =
+        leftOrRightDataFromPair != null ? currentDataGetter.apply(leftOrRightDataFromPair) : null;
+    if (leftOrRightDataFromPair == null || !equalityPredicate.test(proposedData, currentData)) {
       dataSetter.accept(objectToAddChanges, convertedProposedData);
       return true;
     }
     return false;
   }
+
 
   /**
    * @param proposedData      The data that could be added to objectToAddChanges
@@ -157,10 +173,6 @@ public class ChangeManager<T, L, R> {
 
     return this.compareAndBuildRightChange(proposedData, currentDataGetter, Function.identity(),
         Objects::equals, dataSetter);
-  }
-
-  public T getObjectToAddChanges() {
-    return objectToAddChanges;
   }
 
 }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/updatesubmission/changelist/changes/beneficialowner/psc/CorporateBeneficialOwnerPsc.java
```diff
@@ -11,7 +11,7 @@ public class CorporateBeneficialOwnerPsc extends Psc {
   @JsonProperty("corporateName")
   private String corporateName;
   @JsonProperty("companyIdentification")
-  private CompanyIdentification companyIdentification;
+  private CompanyIdentification companyIdentification = new CompanyIdentification();
```
n-r-b, line 14: Since we're not currently initialising objects in our models, can this be done in the service instead?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
```diff
@@ -23,6 +23,10 @@
    * @param objectToAddChanges The object containing the data that has been changed
    */
   public ChangeManager(T objectToAddChanges, Pair<L, R> pair) {
+    if (objectToAddChanges == null) {
+      throw new IllegalArgumentException("objectToAddChanges cannot be null");
+    }
+
```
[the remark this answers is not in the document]
RobertBromfield: Currently, that is being done in the service but I can also do that here 

Do you want an Exception thrown for that as well?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
```diff
@@ -52,21 +56,28 @@ private <P, C, S, U> boolean compareAndBuildChange(
...
-      dataSetter.accept(objectToAddChanges, convertedProposedData);
-      return true;
+    if (!isEmpty(convertedProposedData)) {
+      C currentData = null;
+      if (currentDataGetter != null && leftOrRightDataFromPair != null) {
+        currentData = currentDataGetter.apply(leftOrRightDataFromPair);
+      }
+
+      if (leftOrRightDataFromPair == null || (equalityPredicate != null && !equalityPredicate.test(
+          proposedData, currentData))) {
```
AndreKramer7654, line 71: Should we be comparing the convertedProposedData to the currentData? 
proposedData is in terms of BO/MOP model while current data and changed/setted are in terms of PCSs/Officers?
Or are we comparing apples to oranges?
If not then only need to convert if setting so something does not seem to line up here.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
```diff
@@ -52,21 +56,28 @@ private <P, C, S, U> boolean compareAndBuildChange(
       BiPredicate<P, C> equalityPredicate,
       BiConsumer<T, S> dataSetter) {
 
-    var convertedProposedData = proposedConverter.apply(proposedData);
-
-    if (leftOrRightDataFromPair == null && objectToAddChanges == null) {
-      return false;
+    S convertedProposedData = null;
+    if (proposedConverter != null && proposedData != null) {
```
n-r-b, line 60: I feel the below is becoming way too complex, is there a way to do the null checks and return false up front?

Example:

```
   if (proposedConverter == null || (proposedData == null || isEmpty(proposedData))) {
      return false;
    }
```

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
RobertBromfield: So the converter converts the data found in the public/private data into something that will be accepted in the model

The equalityPredicate checks to see if the data from the proposedData (i.e. The submission DTOs) is the same as that is the public/private data 

It is possible and I'm pretty sure does happen a number of times that the proposedData DataType is different to both the currentData DataType and the Setter DataType 

Hence why the converted data isn't used here

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
RobertBromfield: Yeah you're right 

Though `proposedData == null` isn't needed as it's tested in isEmpty()

n-r-b: That was just an example, there may be other checks to add to this. The same for the block where we throw an exception

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
```diff
@@ -52,24 +59,37 @@ private <P, C, S, U> boolean compareAndBuildChange(
...
-        !isEmpty(convertedProposedData) &&
-            (leftOrRightDataFromPair == null ||
-                !equalityPredicate.test(proposedData,
-                    currentDataGetter.apply(leftOrRightDataFromPair)))
-    ) {
-      dataSetter.accept(objectToAddChanges, convertedProposedData);
-      return true;
+    var convertedProposedData = proposedConverter.apply(proposedData);
+
+    if (!isEmpty(convertedProposedData)) {
```
n-r-b, line 77: Is this check needed since the `proposedData` is not empty at this point?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManager.java
```diff
@@ -52,24 +59,37 @@ private <P, C, S, U> boolean compareAndBuildChange(
...
+      C currentData =
+          leftOrRightDataFromPair != null ? currentDataGetter.apply(leftOrRightDataFromPair) : null;
+      if (leftOrRightDataFromPair == null || !equalityPredicate.test(proposedData, currentData)) {
+        if (dataSetter != null) {
```
n-r-b, line 81: Can this check be moved up? `if (dataSetter == null) { throw new ... `

src/test/java/uk/gov/companieshouse/overseasentitiesapi/utils/ChangeManagerTest.java
```diff
@@ -74,30 +89,179 @@ void testCompareAndBuildPublicChangeNullCurrent() {
...
+
+    Exception exception = assertThrows(IllegalArgumentException.class, () ->
+        changeManager.compareAndBuildLeftChange("newValue",
+            String::new,
+            a -> a,
+            null,
+            Dummy::setField)
+    );
+
+    String expectedMessage = "nvalid input parameter: equalityPredicate is null.";
```
n-r-b, line 258:
```suggestion
    String expectedMessage = "Invalid input parameter: equalityPredicate is null.";
```

bbell-companieshouse (approved):

PR #266, opened 2023-06-06
fix/UAR-580-Fixed-Pair-Null-Issue

RobertBromfield:
### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/BeneficialOwnerChangeServiceTest.java
+++ b/src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/BeneficialOwnerChangeServiceTest.java
@@ -4,15 +4,19 @@ import static com.mongodb.assertions.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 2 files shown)


bbell-companieshouse (approved):

src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/BeneficialOwnerChangeServiceTest.java
```diff
@@ -269,21 +291,65 @@ void testCollateAllBeneficialOwnerChanges() {
...
+    PrintStream standardOut = System.out;
+    ByteArrayOutputStream outputStreamCaptor = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(outputStreamCaptor));
+    
+    List<Change> result = beneficialOwnerChangeService.collateBeneficialOwnerChanges(
+        publicPrivateBo, overseasEntitySubmissionDto);
+
+    System.setOut(standardOut);
+
+    assertEquals("", outputStreamCaptor.toString());
```
n-r-b, line 332: What's the purpose of this? We just want to assert the collateBeneficialOwnerChanges method returns an empty list of changes when the BOs have no chips reference

RobertBromfield: The purpose was to make sure that it wasn't doing any log messages 

With this one, we want the logic that gets rid of the BOs to happen in the stream and not further down.

If there for whatever reason is a Chips Reference it doesn't recognise it will return null but also there will be a log message 

By checking there are no log messages we can check that this is happening where it is supposed to be.

n-r-b (approved):

PR #292, opened 2023-06-22
UAR:790 Changing Statement date logic to correct format

azafarCH:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-790

### Change description

Changed getMonth to getMonth value so we are receiving numbered month, also formatting day and month to include 0 if single digit.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmission.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmission.java
@@ -139,10 +139,10 @@ public class PopulateUpdateSubmission {
                 .ifPresent(filingDate -> {
                     var filingForDate = new FilingForDate();
                     Optional.of(filingDate).map(LocalDate::getDayOfMonth)
-                            .map(Objects::toString)
+                            .map(value -> String.format("%02d",value))
                             .ifPresent(filingForDate::setDay);
-                    Optional.of(filingDate).map(LocalDate::getMonth)
-                            .map(Objects::toString)
+                    Optional.of(filingDate).map(LocalDate::getMonthValue)
+                            .map(value -> String.format("%02d",value))
                             .ifPresent(filingForDate::setMonth);
                     Optional.of(filingDate).map(LocalDate::getYear)
                             .map(Objects::toString)
--- a/src/test/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmissionTest.java
+++ b/src/test/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmissionTest.java
@@ -136,8 +136,4 @@ class PopulateUpdateSubmissionTest {
         assertEquals("France", dueDiligenceCorrespondenceAddress.getCountry());
         assertEquals("NOW 3RE", dueDiligenceCorrespondenceAddress.getPostCode());
     }
-
-    public static String toStringFormat(int localDate){
-        return String.format("%02d", localDate);
-    }
 }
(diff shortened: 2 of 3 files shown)


src/test/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmissionTest.java
```diff
@@ -89,11 +89,11 @@ private void checkStatementDetailsArePopulated(UpdateSubmission updateSubmission
 
     private void checkFilingForDataDetailsArePopulated(UpdateSubmission updateSubmission) {
         assertNotNull(updateSubmission.getFilingForDate());
-        assertEquals(String.valueOf(LOCAL_DATE_TODAY.getDayOfMonth()),
+        assertEquals(toStringDate(LOCAL_DATE_TODAY.getDayOfMonth()),
```
jkayyala, line 92: I would suggest creating a date which can be single digits (ex:1st Feb 2022) so that we can test that day and month is converted 01 and 02 rest

rafahop: Agree with Jitin, let's use a fixed value

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmission.java
```diff
@@ -151,5 +153,5 @@ private void populateFilingForDate(
                             .ifPresent(updateSubmission::setFilingForDate);
                 });
     }
-
+    Function <Integer, String> toStringDate = value -> String.format("%02d",value);
```
RobertBromfield, line 156: As an instance variable, this should probably be at the top of the class 

That said as it's only used in this method I think it should be fine to just define it in `populateFilingForDate()` or even just put the lambda expression in the `map` itself

rafahop, line 156: Can this be a local variable? And maybe find a better name as it is not a full date what it formats?

RobertBromfield (dismissed):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmission.java
RobertBromfield: Oops snap haha

RobertBromfield:
That approval was an accident please ignore it

src/test/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmissionTest.java
azafarCH: I think unfortunately the fields/elements are populated by the PopulateUpdateSubmission class as well, so would alter consistency, but in this case, with the month we are going from 6 to 06, is this okay?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmission.java
azafarCH: ooh yeah, gonna insert the lambda exp straight in, thanks!

src/test/java/uk/gov/companieshouse/overseasentitiesapi/utils/PopulateUpdateSubmissionTest.java
azafarCH: nvm, will change, shouldn't matter much

AndreKramer7654 (approved):

rafahop (approved):

jkayyala (approved):

PR #293, opened 2023-06-23
fix/UAR-591 Fixing Address Equals

RobertBromfield:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-591

### Change description

This PR is to make sure that the comparison of the two addresses will still be equal in the following circumstances 

| Circumstance                                 | Example of 1st Address Line | Example of 2nd Address Line | equals() result |
| -------------------------------------------- | --------------------------- | --------------------------- | --------------- |
| The capitilisation is different              | "123 MAIN STREET"           | "123 Main Street"           | True            |
| The spacing is inconsistent                  | "123 &nbsp; &nbsp; &nbsp; Main &nbsp; &nbsp; Street" | "123 Main Street"           | True            |
|                                              | "123 Ma in Street"          | "123 Main Street"           | False           |
|                                              | " 123 Main Street "         | "123 Main Street"           | True            |
| One line is a blank string the other is null | ""                          | `null`                      | True            |
| One line is a space the other is null        | " "                         | `null`                      | True            |
| One line is a space the other blank string   | " "                         | ""                          | True            |

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/updatesubmission/changelist/commonmodels/Address.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/updatesubmission/changelist/commonmodels/Address.java
@@ -1,13 +1,15 @@
 package uk.gov.companieshouse.overseasentitiesapi.model.updatesubmission.changelist.commonmodels;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
-
-import java.util.Objects;
-
 import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Objects;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


bantao-ch (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/updatesubmission/changelist/commonmodels/Address.java
```diff
@@ -48,6 +50,18 @@ public class Address {
     @JsonProperty("country")
     private String country;
 
+    private static String normalise(String value) {
```
mgibson-ch, line 53: Doesn't StringUtils.normalizeSpace return null if null is passed? Not sure of the point of this method.

RobertBromfield: You're right in that it will return null if the value is null so I've gotten rid of the null check 

However, we still need a way to ensure that it is an empty string after it has gone through `StringUtils.normalizeSpace()` then it is equal to null 

Because of that, I'm going to keep this method but without the null check

PR #296, opened 2023-06-27
Add ROE Update Service Permissions

bantao-ch:
### JIRA link
            [UAR-96](https://companieshouse.atlassian.net/browse/UAR-96)
            [UAR-95](https://companieshouse.atlassian.net/browse/UAR-95)
            https://companieshouse.atlassian.net/browse/UAR-300
          

### Change description

        Add ROE Update Service permissions to the Service.

### Work checklist

- [ Y ] Tests added where applicable

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
@@ -1,6 +1,8 @@
 package uk.gov.companieshouse.overseasentitiesapi.interceptor;
 
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
@@ -19,10 +21,12 @@ import uk.gov.companieshouse.api.util.security.SecurityConstants;
 import uk.gov.companieshouse.api.util.security.TokenPermissions;
 import uk.gov.companieshouse.overseasentitiesapi.utils.ApiLogger;
 
+import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.COMPANY_NUMBER_KEY;
 import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.ERIC_REQUEST_ID_KEY;
+import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.ERIC_AUTHORISED_TOKEN_PERMISSIONS_HEADER;
 import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.TRANSACTION_ID_KEY;
 
-@Component
+@Component("RoeUserAuthenticationInterceptor")
 public class UserAuthenticationInterceptor implements HandlerInterceptor {
 
     /**
@@ -48,17 +52,31 @@ public class UserAuthenticationInterceptor implements HandlerInterceptor {
             return true;
         }
 
+
         // TokenPermissions should have been set up in the request by TokenPermissionsInterceptor
         final var tokenPermissions = getTokenPermissions(request)
                 .orElseThrow(() -> new IllegalStateException("UserAuthenticationInterceptor - TokenPermissions object not present in request"));
 
+        boolean isRoeUpdateJourneyRequest = isRequestROEUpdate(request);
+
         // Check the user has the company_incorporation=create permission
         boolean hasCompanyIncorporationCreatePermission = tokenPermissions.hasPermission(Key.COMPANY_INCORPORATION, Value.CREATE);
+
+        // Check the use has the company_oe_annual_update=create permission for ROE Update
+        boolean hasCompanyOEAnnualUpdateCreatePermission = tokenPermissions.hasPermission(Key.COMPANY_OE_ANNUAL_UPDATE, Value.CREATE);
 
         var authInfoMap = new HashMap<String, Object>();
         authInfoMap.put(TRANSACTION_ID_KEY, transactionId);
         authInfoMap.put("request_method", request.getMethod());
+        authInfoMap.put("is_roe_update_journey_request", isRoeUpdateJourneyRequest);
+        authInfoMap.put("has_company_oe_annual_update_create_permission", hasCompanyOEAnnualUpdateCreatePermission);
         authInfoMap.put("has_company_incorporation_create_permission", hasCompanyIncorporationCreatePermission);
+
+        if (isRoeUpdateJourneyRequest && hasCompanyOEAnnualUpdateCreatePermission) {
+                ApiLogger.debugContext(reqId, "UserAuthenticationInterceptor authorised with company_oe_annual_update=create permission",
+                        authInfoMap);
+                return true;
+        }
 
         if (hasCompanyIncorporationCreatePermission) {
             ApiLogger.debugContext(reqId, "UserAuthenticationInterceptor authorised with company_incorporation=create permission",
@@ -74,4 +92,25 @@ public class UserAuthenticationInterceptor implements HandlerInterceptor {
     protected Optional<TokenPermissions> getTokenPermissions(HttpServletRequest request) {
         return AuthorisationUtil.getTokenPermissions(request);
     }
+
+    protected boolean isRequestROEUpdate(HttpServletRequest request) {
+        final Map<String, List<String>> privileges = getERICTokenPermissions(request);
+        return privileges.containsKey(COMPANY_NUMBER_KEY);
+    }
+
+    private Map<String, List<String>> getERICTokenPermissions(HttpServletRequest request) {
+        String tokenPermissionsHeader = request.getHeader(ERIC_AUTHORISED_TOKEN_PERMISSIONS_HEADER);
+
+        Map<String, List<String>> permissions = new HashMap<>();
+
+        if (tokenPermissionsHeader != null) {
+            for (String pair : tokenPermissionsHeader.split(" ")) {
+                String[] parts = pair.split("=");
+                permissions.put(parts[0], Arrays.asList(parts[1].split(",")));
+            }
+        }
+
+        return permissions;
+    }
+
 }
--- a/src/test/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptorTest.java
+++ b/src/test/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptorTest.java
@@ -53,7 +53,7 @@ class UserAuthenticationInterceptorTest {
     void testInterceptorReturnsTrueWhenRequestHasCorrectTokenPermission() {
         MockHttpServletResponse mockHttpServletResponse = new MockHttpServletResponse();
         Object mockHandler = new Object();
-
+        ericTokenSet(false);
         when(mockHttpServletRequest.getAttribute(TOKEN_PERMISSIONS)).thenReturn(mockTokenPermissions);
         when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_INCORPORATION, Permission.Value.CREATE)).thenReturn(true);
 
@@ -66,8 +66,83 @@ class UserAuthenticationInterceptorTest {
     void testInterceptorReturnsFalseWhenRequestHasIncorrectTokenPermission() {
         MockHttpServletResponse mockHttpServletResponse = new MockHttpServletResponse();
         Object mockHandler = new Object();
+        ericTokenSet(false);
+        when(mockHttpServletRequest.getAttribute(TOKEN_PERMISSIONS)).thenReturn(mockTokenPermissions);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_INCORPORATION, Permission.Value.CREATE)).thenReturn(false);
+
+        var result = userAuthenticationInterceptor.preHandle(mockHttpServletRequest, mockHttpServletResponse, mockHandler);
+        assertFalse(result);
+        assertEquals(HttpServletResponse.SC_UNAUTHORIZED, mockHttpServletResponse.getStatus());
+    }
+
+    @Test
+    void testInterceptorReturnsTrueWhenRequestHasCorrectRoeUpdateTokenPermission() {
+        MockHttpServletResponse mockHttpServletResponse = new MockHttpServletResponse();
+        Object mockHandler = new Object();
+        ericTokenSet(true);
 
         when(mockHttpServletRequest.getAttribute(TOKEN_PERMISSIONS)).thenReturn(mockTokenPermissions);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_INCORPORATION, Permission.Value.CREATE)).thenReturn(false);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_OE_ANNUAL_UPDATE, Permission.Value.CREATE)).thenReturn(true);
+
+        var result = userAuthenticationInterceptor.preHandle(mockHttpServletRequest, mockHttpServletResponse, mockHandler);
+        assertTrue(result);
+        assertEquals(HttpServletResponse.SC_OK,  mockHttpServletResponse.getStatus());
+    }
+
+    @Test
+    void testInterceptorReturnsTrueWhenRequestHasCorrectRegistrationTokenPermission() {
+        MockHttpServletResponse mockHttpServletResponse = new MockHttpServletResponse();
+        Object mockHandler = new Object();
+        ericTokenSet(false);
+
+        when(mockHttpServletRequest.getAttribute(TOKEN_PERMISSIONS)).thenReturn(mockTokenPermissions);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_OE_ANNUAL_UPDATE, Permission.Value.CREATE)).thenReturn(true);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_INCORPORATION, Permission.Value.CREATE)).thenReturn(true);
+
+        var result = userAuthenticationInterceptor.preHandle(mockHttpServletRequest, mockHttpServletResponse, mockHandler);
+        assertTrue(result);
+        assertEquals(HttpServletResponse.SC_OK,  mockHttpServletResponse.getStatus());
+    }
+
+    @Test
+    void testInterceptorReturnsFalseWhenRequestHasInCorrectRegistrationTokenPermission() {
+        MockHttpServletResponse mockHttpServletResponse = new MockHttpServletResponse();
+        Object mockHandler = new Object();
+        ericTokenSet(false);
+
+        when(mockHttpServletRequest.getAttribute(TOKEN_PERMISSIONS)).thenReturn(mockTokenPermissions);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_OE_ANNUAL_UPDATE, Permission.Value.CREATE)).thenReturn(true);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_INCORPORATION, Permission.Value.CREATE)).thenReturn(false);
+
+        var result = userAuthenticationInterceptor.preHandle(mockHttpServletRequest, mockHttpServletResponse, mockHandler);
+        assertFalse(result);
+        assertEquals(HttpServletResponse.SC_UNAUTHORIZED, mockHttpServletResponse.getStatus());
+    }
+
+    @Test
+    void testInterceptorReturnsFalseWhenRequestHasIncorrectRoeUpdateTokenPermission() {
+        MockHttpServletResponse mockHttpServletResponse = new MockHttpServletResponse();
+        Object mockHandler = new Object();
+        ericTokenSet(true);
+
+        when(mockHttpServletRequest.getAttribute(TOKEN_PERMISSIONS)).thenReturn(mockTokenPermissions);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_OE_ANNUAL_UPDATE, Permission.Value.CREATE)).thenReturn(false);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_INCORPORATION, Permission.Value.CREATE)).thenReturn(false);
+
+        var result = userAuthenticationInterceptor.preHandle(mockHttpServletRequest, mockHttpServletResponse, mockHandler);
+        assertFalse(result);
+        assertEquals(HttpServletResponse.SC_UNAUTHORIZED, mockHttpServletResponse.getStatus());
+    }
+
+    @Test
+    void testInterceptorReturnsFalseWhenRequestHasBothIncorrectTokenPermission() {
+        MockHttpServletResponse mockHttpServletResponse = new MockHttpServletResponse();
+        Object mockHandler = new Object();
+        ericTokenSet(true);
+
+        when(mockHttpServletRequest.getAttribute(TOKEN_PERMISSIONS)).thenReturn(mockTokenPermissions);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_OE_ANNUAL_UPDATE, Permission.Value.CREATE)).thenReturn(false);
         when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_INCORPORATION, Permission.Value.CREATE)).thenReturn(false);
 
         var result = userAuthenticationInterceptor.preHandle(mockHttpServletRequest, mockHttpServletResponse, mockHandler);
@@ -86,5 +161,15 @@ class UserAuthenticationInterceptorTest {
         var result = userAuthenticationInterceptor.preHandle(mockHttpServletRequest, mockHttpServletResponse, mockHandler);
         assertTrue(result);
         assertEquals(HttpServletResponse.SC_OK,  mockHttpServletResponse.getStatus());
+    }
+
+    private void ericTokenSet(boolean companyExists) {
+        String tokenValue = null;
+        if (companyExists) {
+            tokenValue = "company_number=00006400";
+        }
+        when(mockHttpServletRequest.getHeader(ERIC_REQUEST_ID_KEY)).thenReturn(REQ_ID);
+        when(mockHttpServletRequest.getHeader(ERIC_IDENTITY_TYPE)).thenReturn(null);
+        when(mockHttpServletRequest.getHeader("ERIC-Authorised-Token-Permissions")).thenReturn(tokenValue);
     }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
         <maven-surefire-plugin.version>2.22.2</maven-surefire-plugin.version>
         <structured-logging.version>1.9.14</structured-logging.version>
-        <api-security-java.version>0.3.9</api-security-java.version>
+        <api-security-java.version>0.4.3</api-security-java.version>
         <log4j.version>2.17.2</log4j.version>
         <org.mapstruct.version>1.5.0.RC1</org.mapstruct.version>
         <rest-service-common-library-version>0.0.5</rest-service-common-library-version>
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/Constants.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/Constants.java
@@ -12,8 +12,14 @@ public class Constants {
     public static final String ERIC_IDENTITY = "ERIC-identity";
     public static final String ERIC_AUTHORISED_USER = "ERIC-Authorised-User";
 
+    public static final String ERIC_AUTHORISED_TOKEN_PERMISSIONS_HEADER
+            = "ERIC-Authorised-Token-Permissions";
+
     // URI path attributes
     public static final String TRANSACTION_ID_KEY = "transaction_id";
+
+    public static final String COMPANY_NUMBER_KEY = "company_number";
+
     public static final String OVERSEAS_ENTITY_ID_KEY = "overseas_entity_id";
 
     // Request attribute names


src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -48,18 +52,32 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
             return true;
         }
 
+
         // TokenPermissions should have been set up in the request by TokenPermissionsInterceptor
         final var tokenPermissions = getTokenPermissions(request)
                 .orElseThrow(() -> new IllegalStateException("UserAuthenticationInterceptor - TokenPermissions object not present in request"));
 
+        boolean isRoeUpdateJourneyRequest = isRequestROEUpdate(request);
+
```
mwestacott, line 61: I think we should introduce a decision point here, based on whether it's the registration or update journey that is being handled by the API end-point. So use the `isRoeUpdateJourneyRequest` value to decide which permission should be checked and only do specific checks for those journeys. Essentially, an `if` ... `else` block.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +92,25 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
     protected Optional<TokenPermissions> getTokenPermissions(HttpServletRequest request) {
         return AuthorisationUtil.getTokenPermissions(request);
     }
+
+    protected boolean isRequestROEUpdate(HttpServletRequest request) {
+        final Map<String, List<String>> privileges = getERICTokenPermissions(request);
+        return privileges.containsKey(COMPANY_NUMBER_KEY);
```
mwestacott, line 98: In addition to just checking the presence of this key, we think we should check that the value starts with 'OE' and matches the `company_number` stored within the associated CHS Transaction Mongo record.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -48,18 +52,32 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        authInfoMap.put("has_company_oe_annual_update_create_permission", hasCompanyOEAnnualUpdateCreatePermission);
         authInfoMap.put("has_company_incorporation_create_permission", hasCompanyIncorporationCreatePermission);
 
+        if (isRoeUpdateJourneyRequest && hasCompanyOEAnnualUpdateCreatePermission) {
+                ApiLogger.debugContext(reqId, "UserAuthenticationInterceptor authorised with company_oe_annual_update=create permission",
+                        authInfoMap);
+                return true;
+        }
+
         if (hasCompanyIncorporationCreatePermission) {
```
mwestacott, line 81: Just falling into this code block is unsafe - if the user is on the Update journey they should absolutely have the `COMPANY_OE_ANNUAL_UPDATE` permission and if not, the method should fail-fast and return 'false'. I think if we introduce the flow control at line 61 above (see other comment), that should address this.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +109,49 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
         return AuthorisationUtil.getTokenPermissions(request);
     }
+
+    protected boolean isRequestROEUpdate(HttpServletRequest request, String transactionId) {
+        final Map<String, List<String>> privileges = getERICTokenPermissions(request);
+        String companyInScope;
+        if (privileges.containsKey(COMPANY_NUMBER_KEY)) {
+            companyInScope = privileges.get(COMPANY_NUMBER_KEY).get(0);
+            if (StringUtils.isNotEmpty(companyInScope) && companyInScope.startsWith("OE")) {
+                String companyInTransaction = getCompanyInTransaction(request, transactionId);
```
mwestacott, line 119: I don't think this part should be in this method - you're on the ROE update journey if the company number key is populated with a number starting with 'OE'. If it does pass that check but company number doesn't match that against the transaction then that is something else.

I suggest extracting the logic that gets the company number in the transaction and checking it against the number in the header  and calling this on line 87 with an AND clause.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +109,49 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+                if (StringUtils.isNotBlank(companyInTransaction)) {
+                    return companyInScope.equalsIgnoreCase(companyInTransaction);
```
mwestacott, line 121: I would write a specific log error message if the company numbers don't match. It should never happen but if it does, it should be investigated by support.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott, line 119: `companyNumberInTransaction` - both variable and method name.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +109,49 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        var logMap = new HashMap<String, Object>();
+        logMap.put(TRANSACTION_ID_KEY, transactionId);
+        try {
+            final var transaction = transactionService.getTransaction(transactionId, passthroughHeader, reqId);
+            logMap.put(COMPANY_NUMBER_KEY, transaction.getCompanyNumber());
+            ApiLogger.debugContext(reqId, "Transaction successfully retrieved " + transactionId, logMap);
+            return transaction.getCompanyNumber();
+        } catch (Exception e) {
+            ApiLogger.errorContext(reqId, "Error retrieving transaction " + transactionId, e, logMap);
+            return null;
```
mwestacott, line 152: Returning `null` here is not enough/correct. If this happens something has gone wrong with the submission to the end-point (transaction id is wrong) which needs to be reflected with a HTTP 500 error code (see how the `TransactionInterceptor` handles this). So logic/exception handling needs reworking slightly.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +110,52 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
     protected Optional<TokenPermissions> getTokenPermissions(HttpServletRequest request) {
         return AuthorisationUtil.getTokenPermissions(request);
     }
+
+    protected boolean isRequestRoeUpdate(HttpServletRequest request) {
+        final Map<String, List<String>> privileges = getERICTokenPermissions(request);
+        if (privileges.containsKey(COMPANY_NUMBER_KEY)) {
+            companyNumberInScope = privileges.get(COMPANY_NUMBER_KEY).get(0);
+            return StringUtils.isNotEmpty(companyNumberInScope) && companyNumberInScope.startsWith("OE");
+            }
```
AndreKramer7654, line 119: } needs aligning

AndreKramer7654: Should we error if it does not start with OE?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +110,52 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
     protected Optional<TokenPermissions> getTokenPermissions(HttpServletRequest request) {
         return AuthorisationUtil.getTokenPermissions(request);
     }
+
+    protected boolean isRequestRoeUpdate(HttpServletRequest request) {
```
AndreKramer7654, line 114: Maybe this should be called hasOverseasEntityNumber() // or getAndValidateOverseasEntityNumber if we throw  an  error?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +110,52 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        return permissions;
+    }
+
+    private String getCompanyNumberInTransaction(HttpServletRequest request, HttpServletResponse response, String transactionId) {
+        String passthroughHeader = request.getHeader(ApiSdkManager.getEricPassthroughTokenHeader());
+        String reqId = request.getHeader(ERIC_REQUEST_ID_KEY);
+        var logMap = new HashMap<String, Object>();
+        logMap.put(TRANSACTION_ID_KEY, transactionId);
+        try {
+            final var transaction = transactionService.getTransaction(transactionId, passthroughHeader, reqId);
```
AndreKramer7654, line 149: I have some concerns about calling out to an external service within a (user auth) interceptor. Possibly lengthy delay or failure. Alternatively we could cache the company number and do the checks in the request handling.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -17,14 +21,27 @@
...
+import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.ERIC_AUTHORISED_TOKEN_PERMISSIONS;
 import static uk.gov.companieshouse.overseasentitiesapi.utils.Constants.TRANSACTION_ID_KEY;
 
-@Component
+@Component("RoeUserAuthenticationInterceptor")
 public class UserAuthenticationInterceptor implements HandlerInterceptor {
 
+    private final TransactionService transactionService;
+
+    private String companyNumberInScope;
```
mwestacott, line 38: This class-level variable being used to store state is not safe. In a multi-threaded, high-load situation the company number could be overwritten 'mid request' by two different threads.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +110,52 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+    protected boolean isRequestRoeUpdate(HttpServletRequest request) {
+        final Map<String, List<String>> privileges = getERICTokenPermissions(request);
+        if (privileges.containsKey(COMPANY_NUMBER_KEY)) {
+            companyNumberInScope = privileges.get(COMPANY_NUMBER_KEY).get(0);
+            return StringUtils.isNotEmpty(companyNumberInScope) && companyNumberInScope.startsWith("OE");
+            }
+        return false;
+    }
+
+    protected boolean doCompanyNumbersMatch(HttpServletRequest request, HttpServletResponse response, String transactionId) {
```
mwestacott, line 123: I would make this method `private`.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -52,18 +69,37 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        authInfoMap.put("is_roe_update_journey_request", isRoeUpdateJourneyRequest);
 
-        if (hasCompanyIncorporationCreatePermission) {
-            ApiLogger.debugContext(reqId, "UserAuthenticationInterceptor authorised with company_incorporation=create permission",
-                    authInfoMap);
-            return true;
+        if (isRoeUpdateJourneyRequest) {
+            boolean doCompanyNumbersMatch = doCompanyNumbersMatch(request, response, transactionId);
+            // Check the user has the company_oe_annual_update=create permission for ROE Update
+            boolean hasCompanyOeAnnualUpdateCreatePermission = tokenPermissions.hasPermission(Key.COMPANY_OE_ANNUAL_UPDATE, Value.CREATE);
```
AndreKramer7654, line 82: Maybe check token permission first - denial of service protection so that request has to have right scope before expensive tests are done.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott, line 114: This method should be `private` too.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -52,18 +69,37 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+            authInfoMap.put("has_company_oe_annual_update_create_permission", hasCompanyOeAnnualUpdateCreatePermission);
+            authInfoMap.put("roe_update_journey_company_numbers_match", doCompanyNumbersMatch);
+            if (hasCompanyOeAnnualUpdateCreatePermission && doCompanyNumbersMatch) {
+                ApiLogger.debugContext(reqId, "UserAuthenticationInterceptor authorised with company_oe_annual_update=create permission",
+                        authInfoMap);
+                return true;
+            } else {
+                ApiLogger.errorContext(reqId, "UserAuthenticationInterceptor unauthorised for ROE Update Journey ", null, authInfoMap);
+                response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
```
mwestacott, line 91: In the scenario that the call to get the transaction has resulted in an exception, `null` is returned from the `getCompanyNumberInTransaction` method and a 500 status code has been set on the response I think this will be overridden at this point to be SC_UNAUTHORIZED, which is incorrect.

Please could you add a unit test to prove/disprove this theory and rework logic (if required). Thanks.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott, line 114: To avoid storing company number as a class level variable, probably easiest to define a new local, private 'pair' POJO that contains the boolean result and company number and return it from this method.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptorTest.java
```diff
@@ -75,6 +88,75 @@ void testInterceptorReturnsFalseWhenRequestHasIncorrectTokenPermission() {
         assertEquals(HttpServletResponse.SC_UNAUTHORIZED, mockHttpServletResponse.getStatus());
     }
```
mwestacott, line 90: As mentioned above, please add a new test for when the call to retrieve the transaction fails and check that the response status code is 500 (internal server error).

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott: I think interceptor will fail (return false) in this scenario so the remote call will ultimately error, but this should be proven by a unit-test - @bantao-ch can you confirm this please?

Adding a helpful log error message in this scenario might be useful though.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott: In theory the call is fine - we make it in another interceptor anyway - but we could possibly avoid doing it and just reference the transaction that's (hopefully) already retrieved by doing:

        final var transaction = (Transaction) request.getAttribute(TRANSACTION_KEY);

See the `AbstractTransactionStatusInterceptor` class which does this. However, as well as unit-testing this works, you'll need to test with the app, just due to the order in which the interceptors are invoked at run-time.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
bantao-ch: Yes @AndreKramer7654 @mwestacott -- this will return false and fail ...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
bantao-ch: we're really checking if the incoming request is for Update or Reg and branching off from there so I propose leaving it as is.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
bantao-ch: To address this we get the transaction only as a last final step in the verification process only if all other checks have passed.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
bantao-ch: Reworked it it in another way !

mwestacott: I'm happy with this naming.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott: Did you decide not to add the log error message @bantao-ch or was this missed?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -51,19 +66,31 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
-                    authInfoMap);
-            return true;
+        if (isRoeUpdateJourneyRequest) {
+            // Check the user has the company_oe_annual_update=create permission for ROE Update
+            boolean hasCompanyOeAnnualUpdateCreatePermission = tokenPermissions.hasPermission(Key.COMPANY_OE_ANNUAL_UPDATE, Value.CREATE);
+            authInfoMap.put("has_company_oe_annual_update_create_permission", hasCompanyOeAnnualUpdateCreatePermission);
+            if (hasCompanyOeAnnualUpdateCreatePermission) {
+                return doRoeUpdateJourneyChecks(request, response, reqId, companyNumberInScope, transactionId, authInfoMap);
+            }
+        } else if (StringUtils.isEmpty(companyNumberInScope)){
```
mwestacott, line 85: I'm wondering why the `else if` was added? We didn't need this originally when just doing the check for registration, so if not on the update journey then we should apply logic as before (I'd have thought).

bantao-ch: Yeah looks like with the refactor we need this check now ....

src/test/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptorTest.java
```diff
@@ -87,4 +183,17 @@ void testInterceptorReturnsTrueWhenAnApiKeyIsUsed() {
...
+
+    private void ericTokenSet(boolean companyExists) {
+        String tokenValue = null;
+        if (companyExists) {
+            tokenValue = "company_number=" + VALID_OE_NUMBER;
+        }
+        setStubbing(tokenValue);
+    }
+    private void setStubbing(String companyNumberToken) {
```
mwestacott, line 194: Minor: A blank line here would improve readability. There are couple of double of blank lines above too.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
bantao-ch: I've added the company_in_scope to the AuthMap instead would that suffice ?

bantao-ch: This is now added in the new revised logic ---

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
bantao-ch: This has now been fixed -- have reworked the logic .....

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -51,19 +66,31 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
-        // Check the user has the company_incorporation=create permission
-        boolean hasCompanyIncorporationCreatePermission = tokenPermissions.hasPermission(Key.COMPANY_INCORPORATION, Value.CREATE);
+        String companyNumberInScope = getCompanyNumberInScope(request);
+        boolean isRoeUpdateJourneyRequest = isRequestRoeUpdate(companyNumberInScope);
 
         var authInfoMap = new HashMap<String, Object>();
         authInfoMap.put(TRANSACTION_ID_KEY, transactionId);
         authInfoMap.put("request_method", request.getMethod());
-        authInfoMap.put("has_company_incorporation_create_permission", hasCompanyIncorporationCreatePermission);
+        authInfoMap.put("company_number_in_scope", companyNumberInScope);
```
AndreKramer7654, line 75: companyNumberInScope == null ? "" : companyNumberrInScope

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +101,81 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        var logMap = new HashMap<String, Object>();
+        logMap.put(TRANSACTION_ID_KEY, transactionId);
+        try {
+            final var transaction = transactionService.getTransaction(transactionId, passthroughHeader, reqId);
+            logMap.put(COMPANY_NUMBER_KEY, transaction.getCompanyNumber());
+            ApiLogger.debugContext(reqId, "Transaction successfully retrieved " + transactionId, logMap);
+            return transaction.getCompanyNumber();
+        } catch (Exception e) {
+            ApiLogger.errorContext(reqId, "Error retrieving transaction " + transactionId, e, logMap);
+            response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
```
AndreKramer7654, line 175: Will this status not be overwritten with HttpServletResponse.SC_UNAUTHORIZED later?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +101,81 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        if (isValidOeNumber) {
+            boolean doCompanyNumbersMatch = doCompanyNumbersMatch(request, response, companyNumberInScope, transactionId);
+            authInfoMap.put("roe_update_journey_company_numbers_match", doCompanyNumbersMatch);
+            if (doCompanyNumbersMatch) {
+                ApiLogger.debugContext(reqId, "UserAuthenticationInterceptor authorised with company_oe_annual_update=create permission",
+                        authInfoMap);
+                return true;
+            } else {
+                ApiLogger.errorContext(reqId, "UserAuthenticationInterceptor unauthorised for ROE Update Journey as Company Numbers do not match", null, authInfoMap);
+                if (HttpServletResponse.SC_INTERNAL_SERVER_ERROR != response.getStatus()) {
```
AndreKramer7654, line 117: I think it is a bit dangerous to check that the response status is already set. How about doComanyNumberMatch always setting the response code? Then no need to back fill here.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
AndreKramer7654: Ok not but slightly convoluted flow.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +101,81 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        final Map<String, List<String>> privileges = getERICTokenPermissions(request);
+        if (privileges.containsKey(COMPANY_NUMBER_KEY)) {
+            return privileges.get(COMPANY_NUMBER_KEY).get(0);
+        }
+        return null;
+    }
+
+    private boolean doCompanyNumbersMatch(HttpServletRequest request, HttpServletResponse response, String companyNumberInScope, String transactionId) {
+        String companyNumberInTransaction = getCompanyNumberInTransaction(request, response, transactionId);
+        if (StringUtils.isNotBlank(companyNumberInTransaction)) {
```
AndreKramer7654, line 145: Set the error code before returning result of compare.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +101,81 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+            return companyNumberInTransaction.equalsIgnoreCase(companyNumberInScope);
+        }
```
AndreKramer7654, line 147: set response code? And rename method ensureCompanyNumbersMatch()

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott: 👍 thanks.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott: The name of this method - `doCompanyNumbersMatch` - sounds ok to me based on the current implementation as it checks if the numbers match and returns a true/false value. Or are you suggesting that the implementation changes too @AndreKramer7654 ?

AndreKramer7654: yes, just to set the response status here. So check on caller can be avoided. Not renaming method is fine.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +103,82 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        logMap.put(TRANSACTION_ID_KEY, transactionId);
+        try {
+            final var transaction = transactionService.getTransaction(transactionId, passthroughHeader, reqId);
+            logMap.put(COMPANY_NUMBER_KEY, transaction.getCompanyNumber());
+            ApiLogger.debugContext(reqId, "Transaction successfully retrieved " + transactionId, logMap);
+            return transaction.getCompanyNumber();
+        } catch (Exception e) {
+            ApiLogger.errorContext(reqId, "Error retrieving transaction " + transactionId, e, logMap);
+            response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+            return TRANSACTION_SERVER_ERROR;
```
AndreKramer7654, line 179: Instead of returning a special value consider returning Optional<String>. The caller can check for Optional.isEmpty() instead of the special value.

mwestacott: Yes, I agree, using the string return value to influence behaviour is not great - either it contains a company number or it doesn't, so yes, making use of Optional here is a good option I think.

AndreKramer7654: would need Optional<Optiona<String>>

AndreKramer7654 (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +102,83 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        boolean isValidOeNumber = companyNumberInScope.startsWith("OE");
+        if (isValidOeNumber) {
+            boolean doCompanyNumbersMatch = doCompanyNumbersMatch(request, response, companyNumberInScope, transactionId);
+            authInfoMap.put("roe_update_journey_company_numbers_match", doCompanyNumbersMatch);
+            if (doCompanyNumbersMatch) {
+                ApiLogger.debugContext(reqId, "UserAuthenticationInterceptor authorised with company_oe_annual_update=create permission",
+                        authInfoMap);
+                return true;
+            } else {
+                ApiLogger.errorContext(reqId, "UserAuthenticationInterceptor unauthorised for ROE Update Journey as Company Numbers do not match", null, authInfoMap);
```
AndreKramer7654, line 117: could not be matched?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +102,83 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        if (tokenPermissionsHeader != null) {
+            for (String pair : tokenPermissionsHeader.split(" ")) {
+                String[] parts = pair.split("=");
+                permissions.put(parts[0], Arrays.asList(parts[1].split(",")));
+            }
+        }
+        return permissions;
+    }
+
+    private Optional<Optional<String>> getCompanyNumberInTransaction(HttpServletRequest request, HttpServletResponse response, String transactionId) {
```
mwestacott, line 166: This method signature is now a little over-complicated. I've attempted to rework the logic in this [PR](https://github.com/companieshouse/overseas-entities-api/pull/303), which we can merge into this one, if you're happy.

bantao-ch: With the help from @AndreKramer7654 I've reworked the logic to simplify please review ...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -74,4 +102,86 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        try {
+            final var transaction = transactionService.getTransaction(transactionId, passthroughHeader, reqId);
+            logMap.put(COMPANY_NUMBER_KEY, transaction.getCompanyNumber());
+            ApiLogger.debugContext(reqId, "Transaction successfully retrieved " + transactionId, logMap);
+            if (StringUtils.isEmpty(transaction.getCompanyNumber())) {
+                response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+                return Optional.empty();
+            }
+            return Optional.of(transaction.getCompanyNumber());
+        } catch (Exception e) {
```
AndreKramer7654, line 179: This should be catch(ServiceException e)

src/test/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptorTest.java
```diff
@@ -75,6 +88,143 @@ void testInterceptorReturnsFalseWhenRequestHasIncorrectTokenPermission() {
...
+    }
+
+    @Test
+    void testInterceptorReturns500StatusWhenTransactionCallFails() {
+        MockHttpServletResponse mockHttpServletResponse = new MockHttpServletResponse();
+        Object mockHandler = new Object();
+        ericTokenSet(true);
+
+        when(mockHttpServletRequest.getAttribute(TOKEN_PERMISSIONS)).thenReturn(mockTokenPermissions);
+        when(mockTokenPermissions.hasPermission(Permission.Key.COMPANY_OE_ANNUAL_UPDATE, Permission.Value.CREATE)).thenReturn(true);
```
AndreKramer7654, line 220: add:
   when(transactionService.getTransaction(any(), any(), any())).thenThrow(new ServiceException("Transaction Service Mock"));

AndreKramer7654 (approved):

mwestacott (approved):

PR #303, opened 2023-07-04
Refactored logic for getting company number from transaction

mwestacott:


dlloyd1-ch (approved):
Looks good to me.

bantao-ch (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -172,9 +175,9 @@ private Optional<Optional<String>> getCompanyNumberInTransaction(HttpServletRequ
             final var transaction = transactionService.getTransaction(transactionId, passthroughHeader, reqId);
             logMap.put(COMPANY_NUMBER_KEY, transaction.getCompanyNumber());
             ApiLogger.debugContext(reqId, "Transaction successfully retrieved " + transactionId, logMap);
-            return Optional.of(Optional.ofNullable(transaction.getCompanyNumber()));
+            return Optional.of(transaction.getCompanyNumber());
```
AndreKramer7654, line 178: Can tranaction.getCompanyNumber() not return null? 
In which case Optiona.of(null) would throw

bantao-ch: Yeah that's Martin's approach to let it throw and deal with it as a 500 status

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -172,9 +175,9 @@ private Optional<Optional<String>> getCompanyNumberInTransaction(HttpServletRequ
...
         } catch (Exception e) {
-            ApiLogger.errorContext(reqId, "Error retrieving transaction " + transactionId, e, logMap);
+            ApiLogger.errorContext(reqId, "Error retrieving company number for transaction " + transactionId, e, logMap);
```
AndreKramer7654, line 180: Was existing code but "catch (Exception e)" will swallow any exception.
Is the intention to let line 178 error though here with a server error? A bit convoluted and should be UNAUTHORIZED if no company number?

AndreKramer7654: A bad actor can use a registration transaction that has no company number. Should that not result in UNAUTHORISED?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott: Yes, exactly.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott: Agreed that a bad actor could try to perform an update with a transaction that has no company number  even though the company number in the header starts with 'OE', but in this scenario we still return 'false' from the `preHandle` method. It's just the HTTP code is INTERNAL SERVER ERROR and not UNAUTHORIZED. Outcome for the user/client is the same though - call to the API end-point will fail.

If we think setting UNAUTHORIZED in this scenario is so important then a larger part of the class logic will need reworking - work that I didn't want to take on at this stage in the original PR's lifecycle.

AndreKramer7654: I'd prefer to have anything security related flagged appropriately. Brian's PR does that and avoids tipping up on NRE but at the cost of one extra level of Optional safety.

AndreKramer7654: Possibly we could throw a SercurtityException of some kind if we don't find a company number and catch that and return UUNAUTHORIZED. But effectively the same as the extra null checking.

mwestacott: Yes, that's an option. Even define a private, custom `CompanyNumberNotFoundException` as an inner-class, throw that and catch it.

PR #309, opened 2023-07-06
ROE Update Service Permissions New Implementation

bantao-ch:
### JIRA link

           https://companieshouse.atlassian.net/browse/UAR-891 

### Change description

            New Implementation for the ROE Update Permissions See Jira Link for Details.

### Work checklist

- [ Y] Tests added where applicable

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
@@ -83,7 +83,7 @@ public class UserAuthenticationInterceptor implements HandlerInterceptor {
         var authInfoMap = new HashMap<String, Object>();
         authInfoMap.put(TRANSACTION_ID_KEY, transactionId);
         authInfoMap.put("request_method", request.getMethod());
-        boolean isRoeUpdateJourneyRequest = false;
+        var isRoeUpdateJourneyRequest = false;
         if (companyNumberInTransactionOptional.isPresent()) {
             authInfoMap.put("company_number_in_scope", companyNumberInScope);
             authInfoMap.put("company_number_in_transaction", companyNumberInTransactionOptional.get());
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
@@ -96,8 +96,8 @@ public class UserAuthenticationInterceptor implements HandlerInterceptor {
             }
         } else {
             String companyNumberInScope = getCompanyNumberInScope(request);
-            authInfoMap.put("company_number_in_scope", getNotNull(companyNumberInScope));
-            boolean isRoeUpdateJourneyRequest = isRequestRoeUpdate(companyNumberInScope, companyNumberInTransactionOptional);
+            authInfoMap.put("company_number_in_scope", convertNullValuesForLogMapOutput(companyNumberInScope));
+            boolean isRoeUpdateJourneyRequest = isValidRoeUpdateRequest(companyNumberInScope, companyNumberInTransactionOptional);
             authInfoMap.put("is_roe_update_journey_request", isRoeUpdateJourneyRequest);
             if (isRoeUpdateJourneyRequest) {
                 // Check the user has the company_oe_annual_update=create permission for ROE Update
@@ -124,12 +124,10 @@ public class UserAuthenticationInterceptor implements HandlerInterceptor {
         return companyNumberInTransaction.isEmpty();
     }
 
-    private boolean isRequestRoeUpdate(String companyNumberInScope, Optional<String> companyNumberInTransaction) {
+    private boolean isValidRoeUpdateRequest(String companyNumberInScope, Optional<String> companyNumberInTransaction) {
         boolean isValidOeNumberInScope = isValidOeNumber(companyNumberInScope);
-        var isValidOeNumberInTransaction = false;
-        if (companyNumberInTransaction.isPresent()) {
-            isValidOeNumberInTransaction = isValidOeNumber(companyNumberInTransaction.get());
-        }
+        var isValidOeNumberInTransaction = companyNumberInTransaction.isPresent() ? isValidOeNumber(companyNumberInTransaction.get()) : false;
+
         if (isValidOeNumberInScope && isValidOeNumberInTransaction) {
             return companyNumberInScope.equalsIgnoreCase(companyNumberInTransaction.get());
         }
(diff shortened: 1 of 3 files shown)


mwestacott:
Please can we get some more detail into the PR commit comments. A JIRA reference (maybe to the original ticket) and explanation of the issue this fix addresses.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -69,20 +69,38 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
         final var tokenPermissions = getTokenPermissions(request)
                 .orElseThrow(() -> new IllegalStateException("UserAuthenticationInterceptor - TokenPermissions object not present in request"));
         String companyNumberInScope = getCompanyNumberInScope(request);
-        boolean isRoeUpdateJourneyRequest = isRequestRoeUpdate(companyNumberInScope);
```
mwestacott, line 72: I think we should keep the `isRequestRoeUpdate` method as it makes the logic easier to read. It will now become more complex, however, as it will need to check the company number in the transaction.

mwestacott: Since discussed, maybe, rather than try and detect if on the update journey, check if on the registration journey. Makes checks simpler, hopefully.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -68,23 +68,23 @@ public boolean preHandle(@NonNull HttpServletRequest request, @NonNull HttpServl
...
+        Optional<String> companyNumberInTransactionOptional;
+
+        try {
+                companyNumberInTransactionOptional = getCompanyNumberInTransaction(request, transactionId);
+            } catch (ServiceException e) {
+                ApiLogger.errorContext(reqId, "Error retrieving transaction " + transactionId, e, logMap);
+                response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
+                return false;
+        }
+        String companyNumberInScope = getCompanyNumberInScope(request);
```
mwestacott, line 81: `companyNumberInScope` isn't needed by the registration journey so I would defer doing this (together with line 86) until you get into the if...else logic block that deals with update.

I think that might also mean that the `getNotNull` method is no longer required (as company number is then guaranteed to be set to something).

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -104,31 +117,24 @@ protected Optional<TokenPermissions> getTokenPermissions(HttpServletRequest requ
...
+            return companyNumberInScope.equalsIgnoreCase(companyNumberInTransaction.get());
         }
+        return false;
     }
 
-    private boolean isRequestRoeUpdate(String companyNumberInScope) {
-        // Further check for OE number is done in doRoeUpdateJourneyChecks method
-        return StringUtils.isNotEmpty(companyNumberInScope);
+    private boolean isValidOeNumber(String oeNumber) {
+        return StringUtils.isNotEmpty(oeNumber) && oeNumber.startsWith("OE");
```
mwestacott, line 137: I'd be tempted to define a class-level constant named something like `OVERSEAS_ENTITY_COMPANY_NUMBER_PREFIX` for the "OE" string and use that here.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -163,26 +158,28 @@ private Map<String, List<String>> getERICTokenPermissions(HttpServletRequest req
...
+        if (transaction == null) {
             return Optional.empty();
         }
+        logMap.put("company_number_in_transaction", getNotNull(transaction.getCompanyNumber()));
+        ApiLogger.debugContext(reqId, "Transaction successfully retrieved " + transactionId, logMap);
+        return Optional.ofNullable(transaction.getCompanyNumber());
 
     }
 
+    private String getNotNull(String aString) {
```
mwestacott, line 177: As mentioned above, if logic re-organised slightly, I don't think this method is required.

bantao-ch: I've kept this method and using it in 2 places to check for null in the LogMap entry.

mwestacott: I'm just not sure that doing this conversion is correct - with it, if a `null` value or an empty string is in the log map, they both appear as the same thing in the output, e.g.

`"company_number_in_transaction":""`

It could be useful to distinguish between the two. I think just not doing the conversion would be fine (in this case, the null value set in the log map is suppressed from the output), but if continuing to keep this method, I would name it something like `convertNullValuesForLogMapOutput` and return "NULL" rather than the empty string.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -104,31 +120,24 @@ protected Optional<TokenPermissions> getTokenPermissions(HttpServletRequest requ
...
-        } else {
-            ApiLogger.debugContext(reqId, "UserAuthenticationInterceptor unauthorised for ROE Update Journey as Company Number is not a valid OE",
-                    authInfoMap);
-            response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
-            return false;
+    private boolean isRequestRoeRegistration(Optional<String> companyNumberInTransaction) {
+        return companyNumberInTransaction.isEmpty();
+    }
+
+    private boolean isRequestRoeUpdate(String companyNumberInScope, Optional<String> companyNumberInTransaction) {
```
AndreKramer7654, line 127: This should named something like isValidRoeUpdateRequest

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
```diff
@@ -104,31 +120,24 @@ protected Optional<TokenPermissions> getTokenPermissions(HttpServletRequest requ
...
+        boolean isValidOeNumberInScope = isValidOeNumber(companyNumberInScope);
+        var isValidOeNumberInTransaction = false;
```
AndreKramer7654, line 129: var isValidOeNumberInTransaction = companyNumberInTransaction.isPresent()? isValidOeNumber(companyNumberInTransaction.get()) : false;

src/test/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptorTest.java
```diff
@@ -109,7 +125,7 @@ void testInterceptorReturnsTrueWhenRequestHasCorrectRoeUpdateTokenPermission() t
     }
 
     @Test
-    void testInterceptorReturnsFalseWhenRequestHasInCorrectRoeUpdateTokenPermission() {
+    void testInterceptorReturnsFalseWhenRequestHasInCorrectRoeUpdateTokenPermission() throws ServiceException {
```
dlloyd1-ch, line 128: Typo - should be 'Incorrect' instead of 'InCorrect'

src/main/java/uk/gov/companieshouse/overseasentitiesapi/interceptor/UserAuthenticationInterceptor.java
mwestacott: I see that this change has been made, but it's introduced a new code smell that needs addressing.

dlloyd1-ch (approved):

AndreKramer7654 (approved):

PR #327, opened 2023-07-31
uar-946 include entity number when no change

AndreKramer7654:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-946

### Change description

Fix entity name not being passed for no change.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
@@ -52,7 +52,6 @@ import uk.gov.companieshouse.overseasentitiesapi.exception.ServiceException;
 import uk.gov.companieshouse.overseasentitiesapi.exception.SubmissionNotFoundException;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.BeneficialOwnerCorporateDto;
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.BeneficialOwnerIndividualDto;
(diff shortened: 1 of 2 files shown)


bantao-ch (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -231,6 +232,8 @@ private void collectUpdateSubmissionData(UpdateSubmission updateSubmission,
               .addAll(managingOfficerAdditionService.managingOfficerAdditions(submissionDto));
 
       ApiLogger.debug("Updates have been collected", logMap);
+    } else {
+      updateSubmission.setEntityName(submissionDto.getEntityName().getName());
```
rafahop, line 236: Out of curiosity, what is the reason for this to differ in the "change" and "no change" scenario? I would expect the same line and possibly even done in the `populateUpdateSubmission.populate()` method?

AndreKramer7654: It can't change in no change case but we have to get it from the public api in change case. Decided not to get it from the public API in no change case - as there is no need to do so.

AndreKramer7654: we have the fetched public data here but not in populate method (for change case).

azafarCH (approved):

PR #329, opened 2023-08-03
fix/UAR-898 Stray trailing comma on nationality

RobertBromfield:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-898

### Change description

This will remove any trailing commas while making sure that none are added.

The `equalsIndividualNationality` is also more permissive now allowing 
- A list with just a comma `,` and a comma and a space `, ` as a delimiter to both be equal.
- A list with a comma at the beginning or end to be equal to one without it 

### Work checklist

### Merge instructions

```
```


src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ComparisonHelper.java
```diff
@@ -229,14 +230,27 @@ private static String combineAddressLines(String... lines) {
...
+        var normalisedOther = normaliseCommaSeparatedList(other);
+
+        return StringUtils.equalsIgnoreCase(normalisedStr, normalisedOther);
+    }
+
+    private static String normaliseCommaSeparatedList(String list){
+        if(list.contains(",")){
+            return Arrays.stream(list.split(","))
+                    .map(ComparisonHelper::normalise)
+                    .filter(Objects::nonNull)
```
AndreKramer7654, line 249: filter(StringUtils::isNotBlank)

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/NationalityOtherMapping.java
```diff
@@ -1,25 +1,28 @@
...
+     * Converts an array of nationality Strings into a comma-separated list, removing blanks and rogue commas.
+     *
+     * @param nationality Varargs parameter of nationality strings to be processed.
+     * @return A comma-separated string without rogue commas and blank spaces, or an empty string if all inputs are blank.
+     */
+    public static String generateNationalityOtherField(String... nationality) {
+        return Arrays.stream(nationality)
+                .filter(StringUtils::isNotBlank)
+                .map(String::trim)
+                .map(s -> s.replaceAll(",{0,5}$|^,{0,5}", ""))
```
AndreKramer7654, line 22: Please replace complex reg expression with code. Nationality should not include a "," so can just replace(",","")

src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/BeneficialOwnerAdditionServiceTest.java
```diff
@@ -204,7 +204,7 @@ private void assertIndividualBeneficialOwnerDetails(IndividualBeneficialOwnerAdd
         assertEquals("John", individualBeneficialOwnerAddition.getPersonName().getForename());
         assertEquals("Doe", individualBeneficialOwnerAddition.getPersonName().getSurname());
         assertEquals(LocalDate.of(1990, 5, 15), individualBeneficialOwnerAddition.getBirthDate());
-        assertEquals("Irish, Spanish", individualBeneficialOwnerAddition.getNationalityOther());
```
AndreKramer7654, line 207: why change test?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ComparisonHelper.java
RobertBromfield: Normalise already does isNotBlank and returns a null if it comes back true.

If we did `StringUtils::isNotBlank` we'd just be running the same code twice.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/NationalityOtherMapping.java
RobertBromfield: This doesn't split words so I'm a bit confused why that's necessary

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ComparisonHelper.java
RobertBromfield: Well actually it uses `StringUtils::normalizeSpace` then `String::isBlank` but that has the same effect as `StringUtils::isNotBlank`

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/NationalityOtherMapping.java
```diff
@@ -1,25 +1,28 @@
...
+     *
+     * @param nationality Varargs parameter of nationality strings to be processed.
+     * @return A comma-separated string without rogue commas and blank spaces, or an empty string if all inputs are blank.
+     */
+    public static String generateNationalityOtherField(String... nationality) {
+        return Arrays.stream(nationality)
+                .filter(StringUtils::isNotBlank)
+                .map(String::trim)
+                .map(s -> s.replace(",", ""))
+                .map(String::trim)
```
mgibson-ch, line 23: Is another trim required here?

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ComparisonHelper.java
AndreKramer7654: You use isNotBlank below. I now think normalise (the map func)  will never produce a null and only "".

src/test/java/uk/gov/companieshouse/overseasentitiesapi/service/BeneficialOwnerAdditionServiceTest.java
AndreKramer7654: of course. It's the expect - my bad

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ComparisonHelper.java
AndreKramer7654: so we don't need the filter

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/NationalityOtherMapping.java
AndreKramer7654: I'd remove the one on line 21

src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/ComparisonHelper.java
RobertBromfield: Normalise returns null if `text` the result of `text.isBlank()` is true
image

This means that without the filter something like "a, " will become "a,null"


_Looking at that normalise again we should have perhaps done `StringUtils.isBlank()` instead of the null check but that's a tangent_

AndreKramer7654 (approved):

mgibson-ch (approved):

jkayyala (approved):

PR #337, opened 2023-09-07
ROE-2343 Initial changes with new test and example

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2343

### Change description

The current problem:

If a DAO/DTO model field datatype changes (e.g. from `String` to `EntityNameDTO`) that causes a breaking change when reading older JSON model data from Mongo (requiring a new data transformer to be written), this is not picked up by the current `DtoModelChangeTest` unit-test unless the data field is present in the test JSON file. If not present and not detected by developers, it can lead to a later run-time failure, possibly in the LIVE environment.

This PR proposes a solution to that by ensuring that (nearly) all fields present in the DTO/DAO model are present in the current (most recent) JSON test file, thereby ensuring that the model conversions will be exercised and tested.

The code is by no means complete and there may well be an alternative/better approach to the problem. Hence its draft status.

### Work checklist

### Merge instructions

```
```


SteveHicksCH (commented):
Reflection code always seems a bit odd when you read it (because you don't often write this sort of stuff) but I can't think of another way to do what you want here so looks good to me

dlloyd1-ch:
Looks good to me as a POC

PR #369, opened 2023-12-08
UAR-1326 Add a new 'is_remove' field to the OE Mongo data

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1326

### Change description

* Save value to Mongo if present in the received JSON data
* Define a new model version (for Remove changes)
* Update appropriate unit tests

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
@@ -27,7 +27,7 @@ public class OverseasEntitySubmissionDto {
     public static final String MANAGING_OFFICERS_INDIVIDUAL_FIELD = "managing_officers_individual";
     public static final String MANAGING_OFFICERS_CORPORATE_FIELD = "managing_officers_corporate";
     public static final String TRUST_DATA = "trusts";
-
+    public static final String IS_REMOVE_FIELD = "is_remove";
     public static final String UPDATE_FIELD = "update";
 
     @JsonProperty(ENTITY_NAME_FIELD)
@@ -72,6 +72,9 @@ public class OverseasEntitySubmissionDto {
 
     @JsonProperty(UPDATE_FIELD)
     private UpdateDto update;
+
+    @JsonProperty(IS_REMOVE_FIELD)
+    private Boolean isRemove;
 
     @JsonProperty("links")
     private Map<String, String> links;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/utils/Constants.java
```diff
@@ -50,5 +50,5 @@ private Constants() {
 
     public static final String CONCATENATED_STRING_FORMAT = "%s,%s";
     public static final String CONCATENATED_STRING_FORMAT_NO_COMMA = "%s%s";
-    public static final SchemaVersion CURRENT_MONGO_SCHEMA_VERSION = SchemaVersion.VERSION_3_1;
+    public static final SchemaVersion CURRENT_MONGO_SCHEMA_VERSION = SchemaVersion.VERSION_3_2;
```
mattch1, line 53: So we need to change the version here? I thought all that was to do with the entity name field.

mwestacott: We don't _have_ to change the version, because the addition of new fields to the JSON/data model is not a breaking change. The entity name field changes involved converting a new string field (for entity name) into a structure that contained an entity name and that broke with older versions of the JSON.

Nevertheless, good to define a new version at this point I think, as it will identify any new JSON records present in Mongo that were added since the remove changes went live.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dto/OverseasEntitySubmissionDto.java
```diff
@@ -73,6 +73,9 @@ public class OverseasEntitySubmissionDto {
     @JsonProperty(UPDATE_FIELD)
     private UpdateDto update;
 
+    @JsonProperty(IS_REMOVE_FIELD)
+    private Boolean isRemove;
```
dlloyd1-ch, line 77: Wondering if we need a Remove block similar to the Update block. We'll need to store the 'remove' statement later on the journey. Would it be better to keep them together?

mwestacott: Yes, we'll need a remove block later on but was thinking that the fundamental decision/info about whether this was a 'remove' submission sat at a higher-level, so outside of that block. If its value was `true`, then you would expect the remove block to be populated.

dlloyd1-ch: Yes that makes sense.

dlloyd1-ch (approved):
Looks good to me.

PR #370, opened 2023-12-13
Upgrade parent-pom to latest version and use its dependency-check

bcullerton:
Move the dependency-check configuration to parent-pom and remove from application level. Also incorporates new api key parameter which allows downloads from NIST api. NIST are retiring data feeds so the key and therefore dep check up version will become necessary: https://github.com/companieshouse/parent-pom/blob/master/pom.xml#L85

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>uk.gov.companieshouse</groupId>
         <artifactId>companies-house-parent</artifactId>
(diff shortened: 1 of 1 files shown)


mwestacott (approved):
LGTM.

Suggest awaiting one more approval from @dlloyd1-ch or @markpit before merging.

markpit (approved):

PR #372, opened 2023-12-21
update to latest parent pom, remove plugin from application level

bcullerton:
The PR reverts back to using the parent pom for dependency-check analysis. Also updates the parent pom from 2.1.4 to 2.1.5 to extend the time between calls to the NVD api.

### Work checklist

### Merge instructions

```
```


mwestacott (approved):
Looks fine. Unless urgent, would prefer this isn't merged until after the Christmas and New Year break though.

PR #386, opened 2024-04-04
UAR-1461 Partial API Validation of Remove Data

mattch1:
### JIRA link
https://companieshouse.atlassian.net/browse/UAR-1461

### Change description

Added common validation checks already present in update
Introduced check for the update filing date and return a 400 error when it is present.

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
@@ -1,5 +1,6 @@
 package uk.gov.companieshouse.overseasentitiesapi.validation;
 
+import java.time.LocalDate;
 import java.util.Objects;
 
 import org.apache.commons.collections.CollectionUtils;
@@ -9,8 +10,14 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 
 import uk.gov.companieshouse.overseasentitiesapi.model.dto.OverseasEntitySubmissionDto;
+import uk.gov.companieshouse.overseasentitiesapi.model.dto.UpdateDto;
+import uk.gov.companieshouse.overseasentitiesapi.utils.ApiLogger;
 import uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators;
+import uk.gov.companieshouse.overseasentitiesapi.validation.utils.ValidationMessages;
 import uk.gov.companieshouse.service.rest.err.Errors;
+
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.UtilsValidators.setErrorMsgToLocation;
+import static uk.gov.companieshouse.overseasentitiesapi.validation.utils.ValidationUtils.getQualifiedFieldName;
 
 @Component
 public class OverseasEntitySubmissionDtoValidator {
@@ -99,9 +106,19 @@ public class OverseasEntitySubmissionDtoValidator {
     }
 
     private void validateFullRemoveDetails(OverseasEntitySubmissionDto overseasEntitySubmissionDto, Errors errors, String loggingContext) {
-        removeValidator.validateFull(overseasEntitySubmissionDto, errors, loggingContext);
 
-        validateUpdateDetails(overseasEntitySubmissionDto, errors, loggingContext);
+        var updateDto = overseasEntitySubmissionDto.getUpdate();
+
+        if (updateDto == null) {
+            String errorMessage = ValidationMessages.NOT_NULL_ERROR_MESSAGE.replace("%s", OverseasEntitySubmissionDto.UPDATE_FIELD);
+            setErrorMsgToLocation(errors, OverseasEntitySubmissionDto.UPDATE_FIELD, errorMessage);
+            ApiLogger.infoContext(loggingContext, errorMessage);
+        } else {
+            validateUpdateDetails(overseasEntitySubmissionDto, errors, loggingContext);
+            validateFilingDate(updateDto.getFilingDate(), errors, loggingContext);
+        }
+        var removeDto = overseasEntitySubmissionDto.getRemove();
+        removeValidator.validateFull(removeDto, errors, loggingContext);
     }
 
     private void validateFullRegistrationDetails(OverseasEntitySubmissionDto overseasEntitySubmissionDto, Errors errors, String loggingContext) {
@@ -172,9 +189,14 @@ public class OverseasEntitySubmissionDtoValidator {
 
         errors = validatePartialCommonDetails(overseasEntitySubmissionDto, errors, loggingContext);
 
+        var updateDto = overseasEntitySubmissionDto.getUpdate();
+        if (updateDto != null) {
+            validateFilingDate(updateDto.getFilingDate(), errors, loggingContext);
+        }
+
         var removeDto = overseasEntitySubmissionDto.getRemove();
         if (removeDto != null) {
-            removeValidator.validate(overseasEntitySubmissionDto, errors, loggingContext);
+            removeValidator.validate(removeDto, errors, loggingContext);
         }
 
         return errors;
@@ -222,6 +244,17 @@ public class OverseasEntitySubmissionDtoValidator {
     private void validateNoChangeUpdate(OverseasEntitySubmissionDto overseasEntitySubmissionDto, Errors errors, String loggingContext) {
         if (UtilsValidators.isNotNull(overseasEntitySubmissionDto.getPresenter(), OverseasEntitySubmissionDto.PRESENTER_FIELD, errors, loggingContext)) {
             presenterDtoValidator.validate(overseasEntitySubmissionDto.getPresenter(), errors, loggingContext);
+        }
+    }
+
+    private void validateFilingDate(LocalDate filingDate, Errors errors, String loggingContext) {
+        // A filing date should NOT be present for a Remove submission
+        if (filingDate != null) {
+            String qualifiedFieldName = getQualifiedFieldName(OverseasEntitySubmissionDto.UPDATE_FIELD,
+                    UpdateDto.FILING_DATE);
+            String errorMessage = ValidationMessages.SHOULD_NOT_BE_POPULATED_ERROR_MESSAGE.replace("%s", qualifiedFieldName);
+            setErrorMsgToLocation(errors, qualifiedFieldName, errorMessage);
+            ApiLogger.infoContext(loggingContext, errorMessage);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -170,12 +170,11 @@ public Errors validatePartialUpdateDetails(OverseasEntitySubmissionDto overseasE
...
     private Errors validatePartialRemoveDetails(OverseasEntitySubmissionDto overseasEntitySubmissionDto, Errors errors, String loggingContext) {
 
-        // TODO Need to also perform partial validation of the common details (like for update journey). See UAR-1461.
-        //      Probably a call to validatePartialCommonDetails(overseasEntitySubmissionDto, errors, loggingContext)
+        errors = validatePartialCommonDetails(overseasEntitySubmissionDto, errors, loggingContext);
 
         var removeDto = overseasEntitySubmissionDto.getRemove();
         if (removeDto != null) {
-            removeValidator.validate(removeDto, errors, loggingContext);
+            removeValidator.validate(overseasEntitySubmissionDto, errors, loggingContext);
```
markpit, line 177: i think the removeDto is still needed on line 177 instead of overseasEntitySubmissionDto. I think the removeDto refers to the 'remove' section in the overseasEntitySubmissionDto and the removeValidator is the validator for that remove section. Maybe @mwestacott can confirm.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/RemoveValidator.java
```diff
@@ -16,10 +16,12 @@
 @Component
 public class RemoveValidator {
 
-    public Errors validate(RemoveDto removeDto, Errors errors, String loggingContext) {
-
-        // TODO Check that the filing date is 'null' if an Update block is present. See UAR-1461.
-
+    public Errors validate(OverseasEntitySubmissionDto overseasEntitySubmissionDto, Errors errors, String loggingContext) {
+        var updateDto = overseasEntitySubmissionDto.getUpdate();
```
markpit, line 20: I think the validateFilingDate should happen in validatePartialRemoveDetails, and the RemoveValidator should just validate the fields inside the 'remove' section of the json/dto.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott: I think @mattch1 's change is right actually, as the remove validator needs access to the 'higher level' object (`overseasEntitySubmissionDto`) in order to check that the filing date isn't present.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/RemoveValidator.java
mwestacott: validateFull method is already taking the overseasEntitySubmissionDto object though and checking the date, so this approach is already being used.

Very happy to discuss though.

mwestacott: Maybe the compromise is to change the implementation and pass in the RemoveDto and the filing date (but not the overseasEntitySubmissionDto)?

mwestacott: The more I think about it, the more I think you're right @markpit and we need to change this for both the full validation and partial validation implementation. A call perhaps?

markpit: yes ok, i didn't realise the full validation was already doing this.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott: Strike that. See comments below.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/RemoveValidator.java
mattch1: There don't seem to be any satisfactory alternatives to this, are you sure we need to do anything here?

mattch1:
Refactored to exclude any update code in the remove validation.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -99,9 +106,19 @@ private void validateUpdateDetails(OverseasEntitySubmissionDto overseasEntitySub
...
 
-        validateUpdateDetails(overseasEntitySubmissionDto, errors, loggingContext);
+        var updateDto = overseasEntitySubmissionDto.getUpdate();
+
+        if (updateDto == null) {
+            String errorMessage = ValidationMessages.NOT_NULL_ERROR_MESSAGE.replace("%s", OverseasEntitySubmissionDto.UPDATE_FIELD);
+            setErrorMsgToLocation(errors, OverseasEntitySubmissionDto.UPDATE_FIELD, errorMessage);
+            ApiLogger.infoContext(loggingContext, errorMessage);
+        } else {
+            validateUpdateDetails(overseasEntitySubmissionDto, errors, loggingContext);
```
markpit, line 117: the validateUpdateDetails used to run after removeValidator.validateFull and didn't have a updateDto == null check around it, I would move it back underneath the removeValidator.validateFull without the updateDto == null check

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidatorTest.java
```diff
@@ -588,70 +618,132 @@ void testPartialUpdateValidationNoFilingDate() {
...
+        assertFalse(errors.hasErrors());
+        verify(removeValidator, times(1)).validateFull(any(), any(), any());
+    }
 
+    @Test
+    void testFullRemoveValidationWithoutTrustsFilingDateIsPresent() {
+        buildOverseasEntitySubmissionDto();
+        overseasEntitySubmissionDto.setIsRemove(true);
+        Errors errors = testFullUpdateRemoveValidationWithoutTrusts();
+
```
markpit, line 641: on 640 the errors returned don't get checked for no errors (it used to be checked in testFullUpdateRemoveValidationWithoutTrusts). I would keep the asserts in the methods where you return the Errors, that way it will assert and also return the Errors as well.

mattch1: We have an assert error check where we expect errors - which we do for all remove scenarios where a filing date (by default) is present (see below)

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mattch1: We require this method call to be in a condition as the first line of the method calls a method on the update dto which will throw an npe.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidatorTest.java
markpit: I wrongly thought that the Errors object returned should be free of errors but in this case it should contain the error that is checked in the lines after this one.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
markpit: I missed the possible npe which could happen on kine 1 of the validateUpdateDetails method so i'm happy with this.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
markpit: We had a call and decided to move the date check out of the remove validator

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/RemoveValidator.java
markpit: Had a call and decided to move the filing date check out of the remove validator.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidatorTest.java
```diff
@@ -772,7 +864,7 @@ void testErrorReportedForMissingUpdateEntityNameField() {
         assertError(qualifiedFieldName, validationMessage, errors);
     }
 
-    private void testFullUpdateRemoveValidationWithoutTrusts() {
+    private Errors testFullUpdateRemoveValidationWithoutTrusts() {
```
mwestacott, line 867: Why did this method signature have to change and the `assertFalse(errors.hasErrors());` line move out of the method?

Same question applies to the methods below.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/RemoveValidatorTest.java
```diff
@@ -79,93 +75,59 @@ void testValidationErrorReportedWhenIsNotProprietorOfLandIsFalse() {
...
         // Then
         String qualifiedFieldName = getQualifiedFieldName(RemoveDto.IS_NOT_PROPRIETOR_OF_LAND_FIELD);
-        String validationMessage = String.format(ValidationMessages.NOT_NULL_ERROR_MESSAGE, qualifiedFieldName);
+        String validationMessage = ValidationMessages.NOT_VALID_ERROR_MESSAGE.replace("%s", qualifiedFieldName);
         assertError(qualifiedFieldName, validationMessage, errors);
     }
 
     @Test
-    void testFullValidationErrorReportedWhenFilingDateIsNotPresentAndIsNotProprietorOfLandIsNull() {
+    void testFullValidationErrorReportedWhenFilingDateIsNotPresentAndRemoveBlockNotPresent() {
```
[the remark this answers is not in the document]
mattch1: Removed filing date from the name.

mwestacott (approved):

markpit (approved):

PR #392, opened 2024-05-10
UAR-1513 Send entity email address in filing data

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/UAR-1513

### Change description

* Entity email address sent now with filing data for a remove submission
* Unit test updated
* Tested with Docker to check that up-to-date email address is sent on both a change and no-change journey

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
@@ -42,7 +42,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.function.Supplier;
 
-import org.apache.commons.lang.NotImplementedException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
@@ -66,6 +65,8 @@ import uk.gov.companieshouse.overseasentitiesapi.utils.PublicPrivateDataCombiner
 public class FilingsService {
 
   private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("d MMMM yyyy");
+
+  private static final String REMOVE_ENTITY_EMAIL_FIELD = "entityEmail";
 
   @Value("${OVERSEAS_ENTITIES_FILING_DESCRIPTION_IDENTIFIER}")
   private String filingDescriptionIdentifier;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
```diff
@@ -181,7 +182,8 @@ private void setFilingApiData(
       ApiLogger.debug("Value of 'isNoChange' flag for Remove is :" + isNoChange, logMap);
       var updateSubmission = new UpdateSubmission();
       collectUpdateSubmissionData(updateSubmission, submissionDto, passThroughTokenHeader, isNoChange, logMap);
-      setRemoveSubmissionData(userSubmission, updateSubmission, isNoChange, submissionDto.getRemove().getIsNotProprietorOfLand(), logMap);
+      setRemoveSubmissionData(userSubmission, updateSubmission, isNoChange, submissionDto.getEntity().getEmail(),
```
markpit, line 185: just wondering if the submissionDto.getEntity().getEmail() would be present for a no change journey? I think it might be but wasn't sure

dlloyd1-ch: I've recently filed no change remove submissions (looking at another issue) and the email address is present in the json sent to the Chips-Filing-Consumer - so should be ok.

dlloyd1-ch (approved):

markpit (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/service/FilingsService.java
dlloyd1-ch: Actually this raises the question, If we're already sending it to the chips-filing-consumer, do we need to include it twice. Could the change be made in the `chips-filing-consumer` to map it from where it currently is: `userSubmission.entity.email`?

markpit: does it not come through though if on a change journey but change something else, ie. it's not in the 'change' section/block

mwestacott: Unresolving. Need to investigate @dlloyd1-ch 's question above and whether the OE API change is actually required.

mwestacott: No, API change not required - we can handle this just in chips-filing-consumer as it's now clear that e-mail address is sent through already, on both a change and no change journey.

mwestacott:
Closed this PR as not needed.

PR #404, opened 2024-07-08
ROE-2567 Persist and validate a new 'has sold land' field

mwestacott:
### JIRA link

https://companieshouse.atlassian.net/browse/ROE-2567

### Change description

* Model and DAO changes to add the new field
* New validation logic added (only run if new API Redis feature flag is true)
* Swagger schema spec updated
* Unit tests added and amended

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
@@ -42,9 +42,8 @@ public class OverseasEntitySubmissionDtoValidator {
     @Value("${FEATURE_FLAG_ENABLE_TRUSTS_WEB_07112022:false}")
     private boolean isTrustWebEnabled;
 
-    @Value("${FEATURE_FLAG_ENABLE_REDIS_REMOVAL_11072024:false}")
+    @Value("${FEATURE_FLAG_ENABLE_REDIS_REMOVAL_15072024:false}")
     private boolean isRedisRemovalEnabled;
-    // check if required based on web PR from Moses!!!
 
     @Autowired
     public OverseasEntitySubmissionDtoValidator(EntityNameValidator entityNameValidator,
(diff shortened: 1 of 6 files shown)


markpit (approved):

lduranteau (approved):

dlloyd1-ch (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -263,4 +267,14 @@ private void validateFilingDate(LocalDate filingDate, Errors errors, String logg
             ApiLogger.infoContext(loggingContext, errorMessage);
         }
     }
+
+    private void validateHasSoldLand(Boolean hasSoldLand, Errors errors, String loggingContext) {
+        if (Boolean.TRUE.equals(hasSoldLand)) {
```
mwestacott, line 272: After discussion with @lduranteau realised that this needs to also check a null value isn't supplied (was originally written thinking this logic had to cope with Update/Remove submissions too, but actually not the case).

markpit: does the 'has_sold_land' appear in all journeys? i though it was missing in some, in which case we would still need to handle a null if the user is not presented with the question/screen.

mwestacott: Doesn't appear on the Update/Remove journeys but then this validation method isn't called on those journeys so we need the flag I think.

mwestacott:
Moved back to draft. Need to decide what's happening with the feature flag for Redis cache removal...

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -42,6 +42,10 @@ public class OverseasEntitySubmissionDtoValidator {
     @Value("${FEATURE_FLAG_ENABLE_TRUSTS_WEB_07112022:false}")
     private boolean isTrustWebEnabled;
 
+    @Value("${FEATURE_FLAG_ENABLE_REDIS_REMOVAL_11072024:false}")
+    private boolean isRedisRemovalEnabled;
+    // check if required based on web PR from Moses!!!
```
mwestacott, line 47: So this may - or may not - be required and used by the checks below.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott: Feature flag now used.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott: Required.

lduranteau (approved):

dlloyd1-ch (approved):

PR #406, opened 2024-07-12
 roe-2570 persist secure-register-filter

lduranteau:
### JIRA link
https://companieshouse.atlassian.net/browse/ROE-2570

### Change description
enable 'Secure-Register-Filter' field to be saved to Mongo and validate appropriately

### Work checklist

### Merge instructions

```
```

### DIFF
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dao/OverseasEntitySubmissionDao.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dao/OverseasEntitySubmissionDao.java
@@ -239,8 +239,8 @@ public class OverseasEntitySubmissionDao {
         return isSecureRegister;
     }
 
-    public void setIsSecureRegister(Boolean hasSecureRegister) {
-        this.isSecureRegister = hasSecureRegister;
+    public void setIsSecureRegister(Boolean isSecureRegister) {
+        this.isSecureRegister = isSecureRegister;
     }
 
     public void setLinks(Map<String, String> links) {
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dao/OverseasEntitySubmissionDao.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dao/OverseasEntitySubmissionDao.java
@@ -72,6 +72,9 @@ public class OverseasEntitySubmissionDao {
     @Field("is_remove")
     private Boolean isRemove;
 
+    @Field("has_secure_register")
+    private Boolean hasSecureRegister;
+
     @Field("trusts")
     private List<TrustDataDao> trusts;
 
@@ -207,6 +210,14 @@ public class OverseasEntitySubmissionDao {
 
     public void setRemove(RemoveDao remove) {
         this.remove = remove;
+    }
+
+    public Boolean getHasSecureRegister() {
+        return hasSecureRegister;
+    }
+
+    public void setHasSecureRegister(Boolean hasSecureRegister) {
+        this.hasSecureRegister = hasSecureRegister;
     }
 
     public Map<String, String> getLinks() {
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dao/OverseasEntitySubmissionDao.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dao/OverseasEntitySubmissionDao.java
@@ -72,8 +72,8 @@ public class OverseasEntitySubmissionDao {
     @Field("is_remove")
     private Boolean isRemove;
 
-    @Field("has_secure_register")
-    private Boolean hasSecureRegister;
+    @Field("is_secure_register")
+    private Boolean isSecureRegister;
 
     @Field("trusts")
     private List<TrustDataDao> trusts;
@@ -213,11 +213,11 @@ public class OverseasEntitySubmissionDao {
     }
 
     public Boolean getHasSecureRegister() {
-        return hasSecureRegister;
+        return isSecureRegister;
     }
 
     public void setHasSecureRegister(Boolean hasSecureRegister) {
-        this.hasSecureRegister = hasSecureRegister;
+        this.isSecureRegister = hasSecureRegister;
     }
 
     public Map<String, String> getLinks() {
--- a/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
+++ b/src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
@@ -42,6 +42,9 @@ public class OverseasEntitySubmissionDtoValidator {
     @Value("${FEATURE_FLAG_ENABLE_TRUSTS_WEB_07112022:false}")
     private boolean isTrustWebEnabled;
 
+    @Value("${FEATURE_FLAG_ENABLE_REDIS_REMOVAL_15072024:false}")
+    private boolean isRedisRemovalEnabled;
+
     @Autowired
     public OverseasEntitySubmissionDtoValidator(EntityNameValidator entityNameValidator,
                                                 EntityDtoValidator entityDtoValidator,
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -123,6 +123,8 @@ private void validateFullRemoveDetails(OverseasEntitySubmissionDto overseasEntit
     }
 
     private void validateFullRegistrationDetails(OverseasEntitySubmissionDto overseasEntitySubmissionDto, Errors errors, String loggingContext) {
+        validateHasSecureRegister(overseasEntitySubmissionDto.getHasSecureRegister(), errors, loggingContext);
```
mwestacott, line 126: Probably doesn't matter, but I put my new 'has sold land' check at the end of the method, after the existing checks.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -263,4 +266,13 @@ private void validateFilingDate(LocalDate filingDate, Errors errors, String logg
             ApiLogger.infoContext(loggingContext, errorMessage);
         }
     }
+
+    private void validateHasSecureRegister(Boolean hasSecureRegister, Errors errors, String loggingContext) {
+        if (Boolean.TRUE.equals(hasSecureRegister) || hasSecureRegister == null) {
```
mwestacott, line 271: In theory, during partial validation, the field may not be supplied, so this shouldn't be an error (I don't think).

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -263,4 +266,13 @@ private void validateFilingDate(LocalDate filingDate, Errors errors, String logg
...
+            String qualifiedFieldName = OverseasEntitySubmissionDto.HAS_SECURE_REGISTER_FIELD + ": " + hasSecureRegister;
```
mwestacott, line 272: Qualified field name should just be the field name (with full path). See implementation for 'has sold land'.

src/test/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidatorTest.java
```diff
@@ -865,6 +866,66 @@ void testErrorReportedForMissingUpdateEntityNameField() throws ServiceException
         assertError(qualifiedFieldName, validationMessage, errors);
     }
 
+    @Test
+    void testNotErrorReportedForHasSecureRegisterFieldSetToFalseForFullValidation() throws ServiceException {
```
mwestacott, line 870: I've now included 'Registration' in the names of the test methods I did and added a couple of tests for Update too. Please take a look.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
lduranteau: Whatever you like, I don't see any difference

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
lduranteau: Yes, you are right it is registration

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
lduranteau: I have added the value so that we know what the error really is (true or null)

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
lduranteau: Removed validation form partial registration

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott: We should know that from the error message. Not always the case, but should be in this situation, as two different messages are required.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott: So we still need to do some validation if it is partial validation - if 'true' is supplied then this is an error. It's just the 'null' check we don't need to do.

dlloyd1-ch (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
mwestacott: This is fine. Will resolve.

src/main/java/uk/gov/companieshouse/overseasentitiesapi/validation/OverseasEntitySubmissionDtoValidator.java
```diff
@@ -263,4 +269,24 @@ private void validateFilingDate(LocalDate filingDate, Errors errors, String logg
...
+
+        String qualifiedFieldName = OverseasEntitySubmissionDto.HAS_SECURE_REGISTER_FIELD;
+
+        if (isFullRegistration && hasSecureRegister == null) {
+            var errorMessage = String.format(ValidationMessages.NOT_NULL_ERROR_MESSAGE, qualifiedFieldName);
+            setErrorMsgToLocation(errors, qualifiedFieldName, errorMessage);
+            ApiLogger.infoContext(loggingContext, errorMessage);
+        }
+
+        if (Boolean.TRUE.equals(hasSecureRegister)) {
```
mwestacott, line 286: If execution of the code has fallen into the error block above (lines 281 to 283 due to a `null` value) the value of `hasSecureRegister` can never be true. You could therefore safely avoid even doing this check by 'stitching' this on to the other `if` block and using an `else` statement.

mwestacott: Yes. Personal preference. This is fine.

mwestacott (approved):

src/main/java/uk/gov/companieshouse/overseasentitiesapi/model/dao/OverseasEntitySubmissionDao.java
```diff
@@ -240,6 +235,14 @@ public void setHasSoldLand(Boolean hasSoldLand) {
         this.hasSoldLand = hasSoldLand;
     }
 
+    public Boolean getIsSecureRegister() {
+        return isSecureRegister;
+    }
+
+    public void setIsSecureRegister(Boolean hasSecureRegister) {
```
mwestacott, line 242: Minor, but parameter name here should be `isSecureRegister`

mwestacott:
The most recent changes look fine but my recent 'has sold land' changes shouldn't be showing in the changed files. I suggest you do a local rebase of this branch against `main` and then push again. Normally they should then disappear.

mwestacott:
> The most recent changes look fine but my recent 'has sold land' changes shouldn't be showing in the changed files. I suggest you do a local rebase of this branch against `main` and then push again. Normally they should then disappear.

These seem to have gone now. Maybe I hadn't refreshed or GitHub was behind. Anyway, all good.

mwestacott (approved):

